PR #4, opened 2022-07-24
Separate field names from results, add a test for the CLI

### LINKED ISSUE #3: Separate field names from results and add tests for command line

marcua:
Resolves #3


tests/cli.rs
```diff
@@ -0,0 +1,90 @@
+use assert_cmd::prelude::*;
+use std::fs;
+use std::process::Command;
+
+#[test]
+fn insert_and_read_rows() -> Result<(), Box<dyn std::error::Error>> {
+    // Delete sqlite files from old runs (it's OK if they don't exist)
```
marcua, line 7:
```suggestion
    // Delete SQLite files from old runs (it's OK if they don't exist)
```

Issue #13, opened 2022-10-23
Authentication & registration


marcua:
Solution outline
- `stacks client register username email@example.com [email|google|github|other OIDC providers]`
    - ensures entityslug and email don't exist or exist only with status=verifying
    - v1: email
         - emails email@example.com saying "To log in, type `stacks client email-confirm email@example.com one-time-token`"
    	 - Token expires in 24h, generated w/ fernet({verson: 1, mode: register, email: email@example.com, entity: entityname}) that encrypts the email address
    - v2: OIDC
       - Likely library: https://docs.rs/openidconnect/latest/openidconnect/#example
       - Example: https://github.com/pka/actix-web-oidc-auth/blob/main/src/main.rs
- `stacks client login username [email|google|github|other OIDC providers]`
	- v1: email
	  - emails email@example.com saying "To log in, type `stacks client email-confirm email@example.com one-time-token`"
	  - Token expires in 24h, generated w/ fernet({verson: 1, mode: login, email: email@example.com, entity: entityname}) that encrypts the email address
- `stacks client email-confirm one-time-token`
    - decrypts fernet
    - if mode == register and status is unverified, transactionally updates entity and creates token
    - updates entity status to verified (if doesn't exist)
     - creates AuthenticationMethod(user, type=email, email=email@example.com)
     - creates token w/ admin permissions, slug `default`. Token data model
		  - unique slug
		  - random ID
		  - created_at datetime
		  - deactivated_at nullable datetime
		  - entity FK
	- for both modes (register and login), returns a bearer token
 	- if multiple active tokens, returns newest one
- v1 bearer token: the random ID, which we look up in the DB on each request
- #39 will use these tokens to validate request permissions

marcua:
Current status: created data models and added CRUD for them to registration endpoint. Added fernet for tokens and lettre for email.

Next up
- [ ] Get email sending to work (need to update test TOML and perhaps add defaults)
- [ ] Create a token with fernet
- [ ] Refactor registration logic to also handle logins

PR #73, opened 2023-04-09
Rename 'create_entity' to 'register'

marcua:
Ahead of #13, renaming this endpoint for clarity.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -59,7 +59,23 @@ impl StacksClient {
             .await
     }
 
-    pub async fn create_entity(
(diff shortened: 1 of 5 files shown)


src/http/client.rs
```diff
@@ -59,7 +59,23 @@ impl StacksClient {
             .await
     }
 
-    pub async fn create_entity(
```
marcua, line 62: These diffs look bigger than they are --- it's just a rename of `create_entity` -> `register` and then moving the function lower in the file to preserve lexicographic order of functions.

PR #212, opened 2023-11-11
Add explicit lifetimes in constants

sofiaritz:
Trying to set up a development setup to start working on a simple UI and cargo suggested this fix :)

Elicit lifetimes in constants will be a compilation error in the future.
See https://github.com/rust-lang/rust/issues/115010


marcua:
Thank you so much @sofiaritz! I learned something new, and ran a `rustup update` :)

PR #222, opened 2023-11-16
Introduce Clippy, apply automated fixes, place endpoints in their own files, and introduce conversion macros

### LINKED ISSUE #21: Add Clippy to check for common errors
Uncomment in in `.github/actions/tests.yml`

sofiaritz:
This branch started to improve the E2E testing experience and maintainability, but ended up becoming an "improvements" branch :p

This PR does the following:
1. Fixes #21 
    * No rules were ignored, the suggestions were applied. The most notable one is this commit: 55b76875b0f5e93a3824c3e89c53247603d5cef1. That commit creates two macros to implement the `TryFrom<i16>` and `FromStr` traits.
    * Clippy now runs in the test workflow.
2. `Display` implementations were removed because they weren't used (at least doesn't seem like that's the case, unit tests and E2E tests passed) and they were a 1-to-1 representation of the `Debug` implementation (which isn't generally desired).


.github/workflows/tests.yml
```diff
@@ -34,8 +34,10 @@ jobs:
     - name: Build
```
marcua, line 34: Discussed offline: I have a somewhat strong preference for pushing as much of the testing, especially when ayb is a young project and subject to lots of internal refactors, into end-to-end testing. To that end, I think
* e2e tests should not be ignored --- they are the key determinant of correctness
* unless the confirm test is testing something the end-to-end test is not, the end-to-end test covers a superset of the code, and the duplicative confirm test should be removed

.gitignore
```diff
@@ -8,6 +8,9 @@
 # Emacs
 *~
 
+# IntelliJ Idea-based IDEs
+.idea/
```
marcua, line 12: Every developer gets to make their political statement in .gitignore. Mine is on line 8 :)

README.md
```diff
@@ -160,3 +160,6 @@ Here's a rough roadmap for the project, with items near the top of the list more
 
 This project has a roadmap and features are added and tested in a certain order. I'm adding a little friction in requiring a discussion/design document for features before submitting a pull request to ensure that I can focus my attention on well-motivated, well-sequenced, and well-understood functionality.
 
+### Running tests
```
marcua, line 163: Might be irrelevant if you agree with the testing-related comment above

src/ayb_db/models.rs
```diff
@@ -2,7 +2,37 @@ use clap::ValueEnum;
...
+                match value {
+                    $($left => Ok($right),)*
+                    _ => Err(()),
+                }
+            }
+        }
+    };
+}
+
+macro_rules! from_str {
```
marcua, line 22: For naming consistency with the macro above

```suggestion
macro_rules! try_from_str {
```

src/http/client.rs
```diff
@@ -37,11 +37,11 @@ impl AybClient {
     ) -> Result<T, AybError> {
         let status = response.status();
         match status {
-            status if status == expected_status => response.json::<T>().await.or_else(|err| {
-                Err(AybError {
+            status if status == expected_status => {
```
marcua, line 40: Sanity-checking that this line is equivalent to the original. We're stripping the `Err` --- is that OK?

src/http/endpoints/confirm.rs
```diff
@@ -0,0 +1,112 @@
+use crate::ayb_db::db_interfaces::AybDb;
```
marcua, line 1: I like that we've pushed each endpoint into its own file. I haven't confirmed line-by-line, but can you confirm that there are no implementation changes in these files, just a cut/paste from `endpoints.rs`?

src/http/endpoints/confirm.rs
```diff
@@ -0,0 +1,112 @@
...
+    let (api_token, token_string) = generate_api_token(&created_entity)?;
+    let _ = ayb_db.create_api_token(&api_token).await?;
+    let returned_token = APIAPIToken {
+        token: token_string,
+    };
+
+    Ok(HttpResponse::Ok().json(returned_token))
+}
+
+#[cfg(test)]
```
marcua, line 73: Relates to the testing-related comment above

src/http/endpoints/testing/mod.rs
```diff
@@ -0,0 +1,27 @@
+use crate::ayb_db::db_interfaces::{connect_to_ayb_db, AybDb};
+use crate::http::structs::{AybConfig, AybConfigAuthentication, AybConfigEmail};
+
+pub async fn test_ayb_database() -> Box<dyn AybDb> {
+    connect_to_ayb_db("sqlite://:memory:".into()).await.unwrap()
+}
+pub fn test_ayb_conf() -> AybConfig {
```
marcua, line 7:
```suggestion
// TODO(sofiaritz): Reuse the default configuration when https://github.com/marcua/ayb/issues/143 is resolved.
pub fn test_ayb_conf() -> AybConfig {
```

marcua: I have a working implementation! Just have to self-code review :)

src/http/structs.rs
```diff
@@ -52,7 +49,10 @@ impl Database {
         Database {
             entity: entity.slug.clone(),
             database: database.slug.clone(),
-            database_type: DBType::from_i16(database.db_type).to_str().to_string(),
+            database_type: DBType::try_from(database.db_type)
+                .expect("unknown database type")
```
marcua, line 53: The nice thing about the pre-macro model is that the error strings were generated in one location, and you could even include the values in the panic. Would it be possible to push the error strings/values into the macros?

tests/utils/mod.rs
```diff
@@ -1,5 +1,5 @@
 use ayb::error::AybError;
-use quoted_printable;
```
marcua, line 2: Wow! I must have left this behind and pushed the logic into the Pythonic SMTP stub server. Thank you!

src/ayb_db/models.rs
```diff
@@ -2,7 +2,37 @@ use clap::ValueEnum;
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use sqlx::FromRow;
-use std::fmt;
+use std::str::FromStr;
+
+macro_rules! try_from_i16 {
```
marcua, line 7: I love these two macros (though wow macros are hard to read :)). Thank you for introducing them! I left one comment related to pushing the error message and value into the macro implementation below.

marcua (changes requested):
Thank you so much for this @sofiaritz. The cleanups, introduction of Clippy, macros, and endpoint cleanup were great. I look forward to your thoughts on the unit tests. You rock!

src/ayb_db/models.rs
sofiaritz: I named the macros according to the trait they implement.
`TryFrom<i16>` &rarr; `try_from_i16`
`FromStr` &rarr; `from_str`

I don't think we should change that, I'll mark the conversation as resolved if you agree

src/http/client.rs
sofiaritz: Yes because it is being replaced with a [`map_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err).
That matching is kinda weird anyways, I'll take a look to it...

src/http/endpoints/confirm.rs
sofiaritz: I did Ctrl+X and Ctrl+V haha
I'll check it later, but it should be exactly the same

.github/workflows/tests.yml
sofiaritz: I've decided to remove the unit tests, I'll mark this as resolved :)

src/http/client.rs
sofiaritz: I've updated the implementation to a more "conventional" one using map and map_err, unless I'm mistaken they are equivalent

sofiaritz:
Addressed most of the comments, I want to see what you think in the others (implementation of conversion traits, etc).
I've finally decided that the unit testing I was implementing didn't really add any benefits.
Ready for review again! :)

src/http/structs.rs
sofiaritz: Fixed with 2b20d875ecb1e8f5f6cceca021c82a4dff63e9d5 :)

src/ayb_db/models.rs
marcua: Sounds great, thank you! :)

marcua (approved):

marcua:
Looks great! Thank you again @sofiaritz ! I learned a lot from this PR! :)

sofiaritz:
Yay!
It's always a pleasure to help :)

PR #224, opened 2023-11-18
Improve cleanup and test creation in E2E testing

### LINKED ISSUE #36: End-to-end test cleanup
Make setup/teardown of e2e tests happen independent of test success/failure.

[See this TODO](https://github.com/marcua/stacks/blob/bf956cf30ea50c3d8bfda4e15e4db090cb45dcf5/tests/e2e.rs#L92)

sofiaritz:
Fixes #36

This PR does two things:
1. Creates three structs: `Cleanup`, `AybServer` and `SmtpServer`
    * `AybServer` and `SmtpServer` are created with their respective `run` function. Once they are dropped, the process in killed; killing the process _may_ fail [silently](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics), or the `Drop::drop` implementation may not be called at all, but in the real world that only happens in a few concrete cases which don't really apply here.
    * `Cleanup` is an [unit struct](https://doc.rust-lang.org/std/keyword.struct.html) whose task is to delete `/tmp/ayb/e2e` on drop.
2. Creates two macros: `ayb_cmd` and `ayb_assert_cmd`.

The `ayb_cmd` macro is equivalent to the following:
```rust
Command::cargo_bin("ayb")?
    .args([...])
    .env("...", ...)
```

The `ayb_assert_cmd` macro is equivalent to the following:
```rust
Command::cargo_bin("ayb")?
    .args([...])
    .env("...", ...)
    .assert()
    .success()
```

Those macros changes an statement like this:
```rust
 Command::cargo_bin("ayb")?
        .args([
            "client",
            "--url",
            server_url,
            "create_database",
            "e2e-first/test.sqlite",
            "sqlite",
        ])
        .env("AYB_API_TOKEN", api_key)
        .assert()
        .success()
        .stdout(format!("{}\n", result));
```

Into this:
```rust
let cmd = ayb_assert_cmd!("client", "--url", server_url, "create_database", "e2e-first/test.sqlite", "sqlite"; {
    "AYB_API_TOKEN" => api_key,
});

cmd.stdout(format!("{}\n", result));
```

### DIFF
--- a/tests/e2e.rs
+++ b/tests/e2e.rs
@@ -1,30 +1,94 @@
 use assert_cmd::prelude::*;
 use regex::Regex;
 use std::fs;
-use std::process::Command;
+use std::process::{Child, Command};
 use std::thread;
 use std::time;
 
 mod utils;
+
+// ayb_cmd!("value1", value2; {
+//     "ENV_VAR" => env_value
+// })
+macro_rules! ayb_cmd {
+    ($($value:expr),+; { $($env_left:literal => $env_right:expr),* $(,)? }) => {
+        Command::cargo_bin("ayb")?
+                .args([$($value,)*])
+                $(.env($env_left, $env_right))*
+    }
+}
+
+// ayb_assert_cmd!("value1", value2; {
+//     "ENV_VAR" => env_value
+// })
+macro_rules! ayb_assert_cmd {
+    ($($value:expr),+; { $($env_left:literal => $env_right:expr),* $(,)? }) => {
+        Command::cargo_bin("ayb")?
+                .args([$($value,)*])
+                $(.env($env_left, $env_right))*
+                .assert()
+                .success()
+    }
+}
+
+struct Cleanup;
+
+impl Drop for Cleanup {
+    fn drop(&mut self) {
+        if let Err(err) = fs::remove_dir_all("/tmp/ayb/e2e") {
+            assert_eq!(format!("{}", err), "No such file or directory (os error 2)")
+        }
+    }
+}
+
+struct AybServer(Child);
+impl AybServer {
+    fn run(db_type: &str) -> Result<Self, Box<dyn std::error::Error>> {
+        let child =
+            ayb_cmd!("server", "--config", &format!("tests/test-server-config-{}.toml", db_type); {
+                "RUST_LOG" => "actix_web=debug",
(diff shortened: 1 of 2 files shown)


tests/e2e.rs
```diff
@@ -1,30 +1,94 @@
...
+
+// ayb_assert_cmd!("value1", value2; {
+//     "ENV_VAR" => env_value
+// })
+macro_rules! ayb_assert_cmd {
+    ($($value:expr),+; { $($env_left:literal => $env_right:expr),* $(,)? }) => {
+        Command::cargo_bin("ayb")?
+                .args([$($value,)*])
+                $(.env($env_left, $env_right))*
+                .assert()
```
marcua, line 29: It might be overengineering or bad practice, but is it worth having `ayb_assert_cmd` call `ayb_cmd` and then to add the final assert/success calls?

tests/e2e.rs
```diff
@@ -1,30 +1,94 @@
...
+        if let Err(err) = fs::remove_dir_all("/tmp/ayb/e2e") {
+            assert_eq!(format!("{}", err), "No such file or directory (os error 2)")
+        }
+    }
+}
+
+struct AybServer(Child);
+impl AybServer {
+    fn run(db_type: &str) -> Result<Self, Box<dyn std::error::Error>> {
+        let child =
```
marcua, line 47: For consistency with `SmtpServer`, should we wrap this in an `Ok(...)` without declaring `let child`?

marcua (changes requested):
Wow @sofiaritz ! I learn something new from each of your PRs! Thank you for these great cleanups.

I left two questions on consistency, and am excited to get this merged!

tests/e2e.rs
sofiaritz: That doesn't compile, a let binding is needed there

tests/e2e.rs
sofiaritz: I could do that, but the tradeoffs (slower compilation times, potential lifetimes issues, etc) are not worth it

tests/e2e.rs
sofiaritz: Done!

sofiaritz:
Done! Ready for review again :)

marcua (approved):
Looks great! Thank you again @sofiaritz !

Issue #216, opened 2023-11-13
A web frontend for `ayb`


marcua:
Some open questions
* How should authentication with the API happen? Do we need session-based auth, or is there a way to rely on the API keys/secrets?
* Can each `ayb` instance host its own web frontend, or is the web frontend project distinct from `ayb`?

sofiaritz:
> I worry a little bit that the loose coupling (...) will add complexity to a new user's mental model as they try to understand ayb

From the perspective of a new user, there are instances with an UI and instances that "only" work with the CLI. A new user doesn't need to understand the relationship between ayb, the CLI, and webAyb.

> The bigger threat to aybWeb and ayb right now is not finishing the v1 of each (...)

hahaha, exactly :)

> We should differentiate between the authentication method (...) from the proof of authentication

Yeah, I got things mixed up there :p

---

About storing and returning the secret I think there are two paths we can take:
1. The Cookie route
    * The servers sets the cookie directly (I like `ayb_sessionid`), that cookie should be both [HttpOnly and with SameSite set to Strict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#security) and potentially encrypted by the server.
    * This means that #214 should still be implemented.
    * This also adds some coupling between frontends and servers because there are now two kinds of authentication proofs: Bearer tokens and cookies (only for web-based clients).
2. The `Authorization` route
     * The server returns a list of authentication proofs ([example](https://git.sofiaritz.com/sofia/gists/src/branch/main/0000-ayb-example-auth-list.md)), then each client decides which one to use. Every client would use the `Authorization` header, maybe with a prefix to help the server identify the kind of token ([example](https://git.sofiaritz.com/sofia/gists/src/branch/main/0001-ayb-example-auth-header.md)).
     * This means #214 doesn't necessarily need to be implemented because the JWT already contains the data.
         * JWT doesn't encrypt the payload, instead it adds a signature that the server can check. This means that the JWT *both* serves as an authentication proof and as a data container.
     * In contrast with the Cookie route, the server only needs to check the `Authorization` header, which also prevents edge cases related to receiving a request with *both* a cookie and a header, and things like that.

I prefer the second one (with or without JWT), but I want to hear your thoughts on those two options :)

sofiaritz:
Unrelated to the current discussion, but related to the web frontend: what is your vision for this web frontend? I just implemented a simple frontend with the endpoints I had available and a few assumptions I made :)

[...]

PR #225, opened 2023-11-20
Add `email.templates.confirm.confirmation_url`

sofiaritz:
This allows server operators to send a URL instead of an `ayb client` command as discussed in #216.

This is the _initial_ templating system. In the future a fully customizable templating system will be added.

Note for implementors: the token is [URI encoded](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) when the URL is generated. Example `email.templates.confirm.confirmation_url`: `https://ayb.sofiaritz.com/auth/confirm?token={token}`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ toml = { version = "0.8.8" }
 tokio = { version = "1.34.0", features = ["macros", "rt"] }
 prefixed-api-key = { version = "0.1.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
-urlencoding = "2.1.3"
+urlencoding = { version = "2.1.3" }
 
 [dev-dependencies]
 assert_cmd = "2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ toml = { version = "0.8.8" }
 tokio = { version = "1.34.0", features = ["macros", "rt"] }
 prefixed-api-key = { version = "0.1.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
+urlencoding = "2.1.3"
 
 [dev-dependencies]
 assert_cmd = "2.0"
--- a/src/email/templating.rs
+++ b/src/email/templating.rs
@@ -1,0 +1,19 @@
+use crate::http::structs::AybConfigEmail;
+
+const CLI_CONFIRM_TMPL: &str = "To complete your registration, type\n\tayb client confirm {token}";
+const WEB_CONFIRM_TMPL: &str = "To complete your registration, visit\n\t <{url}>";
+
+pub fn render_confirmation_template(config: &AybConfigEmail, token: &str) -> String {
+    if let Some(tmpl_conf) = &config.templates {
+        if let Some(confirm_conf) = &tmpl_conf.confirm {
+            return WEB_CONFIRM_TMPL.replace(
+                "{url}",
+                &confirm_conf
+                    .confirmation_url
+                    .replace("{token}", &urlencoding::encode(token)),
+            );
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ toml = { version = "0.8.8" }
 tokio = { version = "1.34.0", features = ["macros", "rt"] }
 prefixed-api-key = { version = "0.1.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
+urlencoding = "2.1.3"
```
marcua, line 32: We can standardize in another PR, but just keeping consistent with the previous lines
```suggestion
urlencoding = { version = "2.1.3" }
```

src/email/templating.rs
```diff
@@ -0,0 +1,19 @@
+use crate::http::structs::AybConfigEmail;
+
+const CLI_CONFIRM_TMPL: &str = "To complete your registration, type\n\tayb client confirm {token}";
+const WEB_CONFIRM_TMPL: &str = "To complete your registration, visit\n\t <{url}>";
+
+pub fn render_confirmation_template(config: &AybConfigEmail, token: &str) -> String {
```
marcua, line 6: What if this lived in email.rs and wasn't a `pub fn`? I suspect it will only ever be used inside of `send_registration_email`, so might not need its own module or public reachability.

src/email/templating.rs
```diff
@@ -0,0 +1,19 @@
+use crate::http::structs::AybConfigEmail;
+
+const CLI_CONFIRM_TMPL: &str = "To complete your registration, type\n\tayb client confirm {token}";
+const WEB_CONFIRM_TMPL: &str = "To complete your registration, visit\n\t <{url}>";
```
marcua, line 4: I feel silly, but what are the `<` and `>` for?

marcua (changes requested):
Looking great! Really tiny comments. Not for this PR, but I added a card to https://github.com/marcua/ayb/projects/1 to document some email settings, including template overrides.

src/email/templating.rs
sofiaritz: This module is not public outside the email module, so this function cannot be used outside the email module

src/email/templating.rs
sofiaritz: Oh just muscle memory from Markdown syntax, I can remove it if you want

sofiaritz:
> I added a card to https://github.com/marcua/ayb/projects/1 to document some email settings, including template overrides.

That's great! Maybe you could also create a tracking issue related to email settings

---

Also, ready for review again :)

src/email/templating.rs
marcua: I think it's good to remove it, since it will go in the email to a client that might not render Markdown

src/email/templating.rs
marcua: That makes sense. My feedback was slightly different, which is to consider having this live as a non-public function in the original `email.rs` file --- having two files for two short and related functions seemed unnecessary :)

marcua:
> Maybe you could also create a tracking issue related to email settings

Yay! Done!

Back over to you! :)

src/email/templating.rs
sofiaritz: Oh alright, I did it that way because in the future the templating logic will be more complex, and thus having a separate file is useful :p

src/email/templating.rs
sofiaritz: Done! :)

sofiaritz:
Ready for review again (pt 2) :)
Also, this will be my last (code) contribution to ayb for the next week or so, if you have any comments feel free to do the changes yourself :)

marcua (approved):
Looks great. Thank you as always @sofiaritz!

PR #231, opened 2023-12-04
Entity retrieval API and `ayb client list` command; Introduce CORS

### LINKED ISSUE #214: Endpoint to retrieve information from an entity
In order to show things like existing databases and to verify that the credentials of a given user are correct, an endpoint to retrieve the information attached to an entity is required.

This endpoint would look something like this:
```bash
$ curl -w "\n" -X GET http://<instance>/v1/entity/sofiaritz -H "authorization: Bearer <valid_token>"

{"entity": "sofiaritz", "databases": ["test.sqlite", "project1.sqlite"]}

$ curl -w "\n" -X GET http://<instance>/v1/entity/sofiaritz -H "authorization: Bearer <invalid_token>"

{"message":"Invalid API token"}
```

sofiaritz:
This PR adds both the entity retrieval API (closes #214) and the `ayb client list` command. The command in E2E tested and this PR also adds support for CORS configuration (related to #216).

The entity retrieval API looks like this:
```bash
$ curl -w "\n" -X GET http://<instance>/v1/entity/sofiaritz -H "authorization: Bearer <valid_token>"

{"slug":"sofiaritz","databases":[{"slug":"amazing-project","database_type":"sqlite"},{"slug":"amazing-project-1","database_type":"sqlite"}]}

$ curl -w "\n" -X GET http://<instance>/v1/entity/sofiaritz -H "authorization: Bearer <invalid_token>"

{"message":"Invalid API token"}
```

The `ayb client list <entity>` command will display the databases owned by the given entity that the current user can query. The output of the command can be formatted using the `--format` argument, just like the `ayb client query` command.


src/http/structs.rs
```diff
@@ -92,6 +99,64 @@ pub struct EntityDatabasePath {
...
+    pub entity: String,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EntityQueryResponse {
+    pub slug: String,
+    pub databases: Vec<EntityDatabase>,
+}
+
+impl EntityQueryResponse {
```
sofiaritz, line 113: I'm not sure whether we should create a trait for _structs that can be converted to tables_, the `generate_table` and `generate_csv` functions are exactly the same as the ones implemented for the query command.

src/http/structs.rs
```diff
@@ -92,6 +99,64 @@ pub struct EntityDatabasePath {
...
+        table.set_titles(Row::new(vec![
+            Cell::new("Database slug"),
+            Cell::new("Type"),
+        ]));
+
+        self.databases
+            .iter()
+            .map(|v| Row::new(vec![Cell::new(&v.slug), Cell::new(&v.db_type)]))
+            .for_each(|c| {
+                table.add_row(c);
```
sofiaritz, line 125: Note to self: inline this function

src/http/endpoints/query_entity.rs
```diff
@@ -0,0 +1,31 @@
+use crate::ayb_db::db_interfaces::AybDb;
+use crate::ayb_db::models::InstantiatedEntity;
+use crate::error::AybError;
+use crate::http::permissions::can_query;
+use crate::http::structs::{EntityDatabase, EntityQueryPath, EntityQueryResponse};
+use crate::http::utils::unwrap_authenticated_entity;
+use actix_web::{get, web};
+
+#[get("/v1/entity/{entity}")]
+pub async fn query_entity(
```
marcua, line 10: So that people don't get confused and think it's related to DB queries

```suggestion
pub async fn entity_details(
```

src/http/client.rs
```diff
@@ -111,6 +111,20 @@ impl AybClient {
             .await
     }
 
+    pub async fn list_databases(&self, entity: &str) -> Result<EntityQueryResponse, AybError> {
```
marcua, line 114: Discussion: the client calls it `list`, but the server calls it `entity_details`/`query_entity`. I get the sense we'll add other details to the endpoint in the future (e.g., name, created at, etc.), so do we want to make the client consistent with that (that it's more of a `details` endpoint than a `list`ing endpoint)? I don't feel strongly now about naming, so this is more so about consistency between client and server naming.

src/http/endpoints/query_entity.rs
```diff
@@ -0,0 +1,31 @@
...
+    let authenticated_entity = unwrap_authenticated_entity(&authenticated_entity)?;
+    let entity_slug = &path.entity;
+    let desired_entity = ayb_db.get_entity_by_slug(entity_slug).await?;
+
+    let databases = ayb_db
+        .list_databases_by_entity(&desired_entity)
+        .await?
+        .into_iter()
+        .filter(|v| can_query(&authenticated_entity, v))
+        .map(Into::<EntityDatabase>::into)
```
marcua, line 24: Silly question, but what's the Into?

src/http/server.rs
```diff
@@ -57,6 +61,22 @@ async fn entity_validator(
...
+    let mut cors = Cors::default().allow_any_header().allow_any_method();
+
+    if ayb_cors
+        .as_ref()
+        .is_some_and(|conf| conf.origin.trim() == "*")
+        || ayb_cors.is_none()
+    {
+        cors = cors.allow_any_origin()
+    } else {
+        cors = cors.allowed_origin(ayb_cors.as_ref().unwrap().origin.trim());
```
marcua, line 74: Are we guaranteed the `unwrap` is safe here?

src/http/server.rs
```diff
@@ -57,6 +61,22 @@ async fn entity_validator(
     }
 }
 
+fn build_cors(ayb_cors: Option<AybConfigCors>) -> Cors {
+    let mut cors = Cors::default().allow_any_header().allow_any_method();
+
+    if ayb_cors
+        .as_ref()
+        .is_some_and(|conf| conf.origin.trim() == "*")
```
marcua, line 69: Brainstorming alternatives --- we could make the CORS configuration required, and default it to `*`, which would simplify some of this logic.

marcua: I don't feel too strongly

src/http/server.rs
```diff
@@ -67,8 +87,11 @@ pub async fn run_server(config_path: &PathBuf) -> std::io::Result<()> {
 
     println!("Starting server {}:{}...", ayb_conf.host, ayb_conf.port);
     HttpServer::new(move || {
+        let cors = build_cors(ayb_conf.cors.clone());
```
marcua, line 90: Worth moving to line 87 so all the config stuff is in one block?

src/http/structs.rs
marcua: Nice idea!

src/http/structs.rs
```diff
@@ -92,6 +99,64 @@ pub struct EntityDatabasePath {
     pub database: String,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EntityQueryPath {
```
marcua, line 103: could call it EntityPath since it could use used in other paths with only Entity slug as well

src/http/structs.rs
marcua: One tiny thought: do we care that we lose the slug in the tabular view? We're only keeping the database list, which is probably OK, but figured I'd point it out to see if we care.

marcua (changes requested):
Looks great! Tiny stuff back over to you! Thank you for adding this endpoint!

Cargo.toml
```diff
@@ -30,6 +30,7 @@ tokio = { version = "1.34.0", features = ["macros", "rt"] }
 prefixed-api-key = { version = "0.1.0", features = ["sha2"]}
```
marcua, line 30: Documentation: Worth also updating the HTTP and CLI examples in the README.md?

src/http/structs.rs
sofiaritz: > do we care that we lose the slug in the tabular view?

Without the database slug what information would be shown to the user?

src/http/endpoints/query_entity.rs
sofiaritz: You're right

src/http/endpoints/query_entity.rs
sofiaritz: I've just realised that that should be `From::from` instead of `Into::<EntityDatabase>::into` :p, I've changed it

[`Into`](https://doc.rust-lang.org/rust-by-example/conversion/from_into.html) is a trait implemented for every struct that also implements `From`. This traits allows you to change `From::from(v)` to `v.into()`, which is useful when chaining function calls.

With the function signature of `from` being: `From::from<T>() -> U`, `into` is the following `U::from(self)`

src/http/structs.rs
sofiaritz: > do we care that we lose the slug in the tabular view?

I've (confusingly) named both the database slugs and the entity slug... slugs in the endpoint. Which one are you referring to?

src/http/server.rs
sofiaritz: I think it should be required, that way the instance owner is aware that CORS is set to `*` by default without having to read documentation. I'll change it

src/http/server.rs
sofiaritz: It is safe because of the `is_none` check above, I'm going to change this anyway

src/http/server.rs
sofiaritz: The `CORS` struct from `actix-cors` doesn't implement `Send`, and the `HttpServer::new` function required the lambda to be `Send`. So that doesn't compile. (Unlike the last time, I tried it before posting the comment, not after :p)

src/http/structs.rs
sofiaritz: Done :)

Cargo.toml
sofiaritz: Done! :)

src/http/structs.rs
sofiaritz: Update to self: I can't because `for_each` cannot return anything :p

src/http/structs.rs
sofiaritz: Also, I think I'll create the trait in a separate PR :)
I don't want to clutter this one a lot :p

sofiaritz:
Oops, wait I have to fix tests... they worked locally though hmm

tests/test-server-config-postgres.toml
```diff
@@ -15,3 +15,6 @@ smtp_password = "the_password"
 [authentication]
 fernet_key = "y3UdMqGh6si7pvQb8wsuW3ryiJcacp0H1QoHUPfsjb0="
 token_expiration_seconds = 3600
+
+[cors]
+origin = "*"
```
marcua, line 20:
```suggestion
origin = "*"

```

Cargo.toml
marcua: Thank you!

marcua:
I updated the PR title to include CORS for the sake of the commit message. This looks GREAT! Thank you for the iterations. One tiny one-line back at you, and then we merge! :)

sofiaritz:
Done :) (I think? The GitHub UI is terrible for new lines and things like that... at least locally the diff shows the new line, feel free to do the change yourself if that is not the case :p)

marcua (approved):

marcua:
Looks great! Thank you again @sofiaritz! :)

Issue #41, opened 2023-02-25
Isolate databases from one-another


marcua:
[A discussion on the SQLite forum](https://sqlite.org/forum/forumpost/0f77e9f96ec58def) that effectively means we want to isolate databases from one another so that one user can't query/edit another user's database.

[Some general background reading](https://fly.io/blog/sandboxing-and-workload-isolation/)

A few options to explore
* Lightweight WASM: https://docs.rs/wasmtime/latest/wasmtime/
* Good old heavyweight Docker: https://docs.rs/docker-command/latest/docker_command/
* Lighter-weight jails: https://nsjail.dev/. Used in this project: https://github.com/windmill-labs/windmill/tree/main/backend/windmill-worker

marcua:
wasmtime compiled on branch https://github.com/marcua/ayb/compare/main...isolate-dbs as of commit https://github.com/marcua/ayb/commit/5225ec8c4eddbdb0a6aa7abac291257992855e8b
* The runner works with a regular rust build
```
marcua@marcua-dev:~/ayb/src/hosted_db_runner$ target/debug/hosted_db_runner 
QueryResult {
    fields: [],
    rows: [],
}
```
* On rusqlite 0.26.1 we get a missing sqlite init import error
```
marcua@marcua-dev:~/ayb/src/hosted_db_runner$ wasmtime target/wasm32-wasi/debug/hosted_db_runner.wasm 
Error: failed to run main module `target/wasm32-wasi/debug/hosted_db_runner.wasm`

Caused by:
    0: failed to instantiate "target/wasm32-wasi/debug/hosted_db_runner.wasm"
    1: unknown import: `env::sqlite3_os_init` has not been defined
```

* On rusqlite 0.30.0, we get a missing floating point library error
```
marcua@marcua-dev:~/ayb/src/hosted_db_runner$ wasmtime target/wasm32-wasi/debug/hosted_db_runner.wasm 
Error: failed to run main module `target/wasm32-wasi/debug/hosted_db_runner.wasm`

Caused by:
    0: failed to instantiate "target/wasm32-wasi/debug/hosted_db_runner.wasm"
    1: unknown import: `env::__extenddftf2` has not been defined
```

marcua:
Trying a different approach, since the newer rusqlite will have different problems than the one on which we built our build script
* There's an [open PR](https://github.com/rusqlite/rusqlite/pull/1332) against a relatively new rusqlite version that's promising
[...]

PR #234, opened 2023-12-09
Limit the number of attached databases to zero and enable defensive mode

sofiaritz:
Related to #41.
This PR limits the number maximum attached databases to zero (none) and enables defensive mode to prevent deliberate corruption of the database.

Refs:
https://www.sqlite.org/c3ref/c_dbconfig_defensive.html
https://www.sqlite.org/lang_attach.html
https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitattached
https://www.sqlite.org/c3ref/limit.html


marcua:
This is perfect! Since I'm refactoring this code, I've copied it verbatim into #234, so I'll close this one and the code will appear once the refactor is done. Thank you for protecting ayb users! :)

sofiaritz:
I forgot to close this myself :p
Really excited for the full DB isolation! :)

PR #235, opened 2023-12-09
Isolate DBs

### LINKED ISSUE #41: Isolate databases from one-another

marcua:
Resolves #41


Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0"
 actix-web = { version = "4.4.0" }
 actix-web-httpauth = { version = "0.8.1" }
 async-trait = { version = "0.1.74" }
+ayb-hosted-db-runner = { path = "src/hosted_db_runner" }
```
sofiaritz, line 14: Maybe creating a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) is a good idea.
There could be three packages: `ayb`, `ayb-hosted-db-runner`, and `ayb-common`.
That way things like AybError can be imported from the `ayb-common` crate. :)

Amazing work!! :)

marcua: Thank you! This is a great recommendation. I'm trying to figure out the right structure for what we need. All we really want is a second binary crate for the runner so that we can wrap it in a `vsjail`. Would it be enough to create two binary crates (`ayb` and `hosted_db_runner`) in the same package and avoiding multiple packages or workspaces entirely, or can we think of some benefit of breaking them apart?

Relatedly, I'm curious if there's a canonical way to reference one binary crate from another so the `ayb` binary can say "run the other binary for the runner" without having to explicitly provide a path to it.

sofiaritz: > Would it be enough to create two binary crates (ayb and hosted_db_runner) in the same package

I have never seen that in the wild tbh. Maybe it's fine, but that's also one of the use cases of workspaces... I would create a top level workspace just to be sure and to also follow the "convention".

This article may be of interest: https://matklad.github.io/2021/08/22/large-rust-workspaces.html

> Relatedly, I'm curious if there's a canonical way to reference one binary crate from another

Not that I am aware of, sadly.
You could embed the runner at compile time with `include_bytes!` and create a copy of the runner at startup time and delete it when the program is stopped, that way the runner is always up-to-date and is also directly tied to the ayb binary.

scripts/build_nsjail.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+git clone https://github.com/google/nsjail.git nsjail-checkout
+cd nsjail-checkout
+make
+mv nsjail ..
+cd ..
+rm -rf nsjail-checkout
+
```
marcua, line 9: Remove

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+        })?
+        .to_str()
+        .ok_or(AybError {
+            message: format!("Invalid DB path {}", absolute_db_path.display()),
+        })?;
+    let tmp_db_path = Path::new("/tmp").join(db_file_name);
+    let db_file_mapping = format!("{}:{}", absolute_db_path.display(), tmp_db_path.display());
+    cmd.args(["--bindmount", &db_file_mapping]);
+
+    // Generate a /local/path/to/ayb_isolated_runner:/tmp/ayb_isolated_runner mapping.
```
marcua, line 65: Add note that this assumes `ayb` and `ayb_isolated_runner` will always be in the same directory

Cargo.toml
```diff
@@ -19,17 +20,25 @@ fernet = { version = "0.2.1" }
...
-tokio = { version = "1.34.0", features = ["macros", "rt"] }
+tokio = { version = "1.34.0", features = ["macros", "process", "rt"] }
 prefixed-api-key = { version = "0.1.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
 
 [dev-dependencies]
 assert_cmd = "2.0"
 assert-json-diff = "2.0.2"
+
+[[bin]]
```
marcua, line 38: I wonder if I need double-brackets

src/bin/ayb_isolated_runner.rs
```diff
@@ -0,0 +1,16 @@
+use ayb::hosted_db::sqlite::query_sqlite;
+use serde_json;
+use std::env;
+use std::path::PathBuf;
+
+fn main() -> Result<(), serde_json::Error> {
```
marcua, line 6: Add comment explaining what this runner is for

Cargo.toml
```diff
@@ -6,6 +6,7 @@ description = "ayb makes it easy to create, host, and share embedded databases l
 homepage = "https://github.com/marcua/ayb"
```
marcua, line 6: Rebase/merge main, resolve conflicts

Cargo.toml
```diff
@@ -6,6 +6,7 @@ description = "ayb makes it easy to create, host, and share embedded databases l
...
 documentation = "https://github.com/marcua/ayb#readme"
 license = "Apache-2.0"
```
marcua, line 8: Add section on isolation and the optional `nsjail` configuration to `README.md`

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+    let mut cmd = tokio::process::Command::new(nsjail);
+
+    cmd.arg("--really_quiet") // log fatal messages only
+        .arg("--iface_no_lo")
+        .args(["--mode", "o"]) // run once
+        .args(["--hostname", "ayb"])
+        .args(["--bindmount_ro", "/lib:/lib"])
+        .args(["--bindmount_ro", "/lib64:/lib64"])
+        .args(["--bindmount_ro", "/usr:/usr"])
+        .args(["--mount", "none:/tmp:tmpfs:size=100000000"]) // TODO(marcua): Restrict disk size more configurably?
```
marcua, line 41: Remove question marks here and on next line :)

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let stdout_reader = BufReader::new(child.stdout.take().unwrap());
+    let stderr_reader = BufReader::new(child.stderr.take().unwrap());
+
+    let output = child.wait_with_output().await?;
+
+    // read stdout and stderr from process
+    let mut stdout = Vec::new();
+    let mut stderr = Vec::new();
+    stdout_reader
```
marcua, line 102: Presumably we'll regret having hard-coded limits on stdout/stderr, add TODO

src/hosted_db/sqlite.rs
```diff
@@ -36,3 +47,35 @@ pub fn run_sqlite_query(path: &PathBuf, query: &str) -> Result<QueryResult, AybE
         rows: results,
     })
 }
+
+pub async fn run_sqlite_query(
```
marcua, line 51: Consider this function name relative to the other one in this file: can they be more distinct?

Cargo.toml
marcua: Phew! I managed to figure out a way without too much hackery and one somewhat safe assumption
* One crate
* Two binaries (`ayb` and ayb_isolated_runner`)
* Since it's the same crate, it's relatively safe to assume both binaries will end up in the same directory (true of build targets as well as where the files will be installed with a `crate install`
* If we assume they are in the same directory, we can call `std::env::current_exe` to get a path to the current binary, which we can then use to create a path to the other binary

Hooray! I'm resolving conflicts, but this PR is largely good for review if you want to take a pass over it. I've self-reviewed and left myself comments that I'll resolve in the coming days, but if you're interested I'd love your opinion! :)

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+    // Generate a /local/path/to/file:/tmp/file mapping.
+    let absolute_db_path = canonicalize(db_path)?;
+    let db_file_name = absolute_db_path
+        .file_name()
+        .ok_or(AybError::Other {
+            message: format!("Invalid DB path {}", absolute_db_path.display()),
+        })?
+        .to_str()
+        .ok_or(AybError::Other {
+            message: format!("Invalid DB path {}", absolute_db_path.display()),
```
marcua, line 59: Consider how to avoid this repetition of the AybError from line 55.

src/http/structs.rs
```diff
@@ -43,6 +48,7 @@ pub struct AybConfig {
     pub email: AybConfigEmail,
     pub web: Option<AybConfigWeb>,
     pub cors: AybConfigCors,
+    pub isolation: Option<AybConfigIsolation>,
 }
 
 impl AybConfig {
```
marcua, line 54: tiny hack: I've checked in a built `nsjail` binary to `tests/nsjail`. In a perfect world I'd containerize it or build it each time from scratch, both of which feel like a decent amount of work given it's just for tests and not actually distributing the binary.

Cargo.toml
```diff
@@ -6,6 +6,7 @@ description = "ayb makes it easy to create, host, and share embedded databases l
 homepage = "https://github.com/marcua/ayb"
 documentation = "https://github.com/marcua/ayb#readme"
```
marcua, line 7: Tests pass locally but fail in GitHub Actions
image

Cargo.toml
sofiaritz: That sounds like some reasonable assumptions and a clean workaround :)
I still don't really like the idea of one crate having two binaries like that... maybe creating a `[[bin]]` entry is a better approach (https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries)

I'll review the PR :)

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+    ]);
+
+    let mut child = cmd
+        .arg("--")
+        .arg("/tmp/ayb_isolated_runner")
+        .arg(tmp_db_path)
+        .arg(query)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdin(Stdio::piped())
```
sofiaritz, line 91: Is piping `stdin` required?

src/hosted_db/sqlite.rs
```diff
@@ -36,3 +47,35 @@ pub fn run_sqlite_query(path: &PathBuf, query: &str) -> Result<QueryResult, AybE
...
+    path: &PathBuf,
+    query: &str,
+    isolation: &Option<AybConfigIsolation>,
+) -> Result<QueryResult, AybError> {
+    match isolation {
+        Some(isolation) => {
+            let result = run_in_sandbox(Path::new(&isolation.nsjail_path), path, query).await?;
+
+            if result.stderr.len() > 0 {
```
sofiaritz, line 60:
```suggestion
            if !result.stderr.is_empty() {
```

src/hosted_db/sqlite.rs
```diff
@@ -36,3 +47,35 @@ pub fn run_sqlite_query(path: &PathBuf, query: &str) -> Result<QueryResult, AybE
...
+                let error: AybError = serde_json::from_str(&result.stderr)?;
+                Err(error)
+            } else if result.status != 0 {
+                Err(AybError::Other {
+                    message: format!(
+                        "Error status from sandboxed query runner: {}",
+                        result.status.to_string()
+                    ),
+                })
+            } else if result.stdout.len() > 0 {
```
sofiaritz, line 70:
```suggestion
            } else if !result.stdout.is_empty() {
```

src/hosted_db/sqlite.rs
```diff
@@ -36,3 +47,35 @@ pub fn run_sqlite_query(path: &PathBuf, query: &str) -> Result<QueryResult, AybE
         rows: results,
     })
 }
+
+pub async fn run_sqlite_query(
+    path: &PathBuf,
+    query: &str,
+    isolation: &Option<AybConfigIsolation>,
+) -> Result<QueryResult, AybError> {
+    match isolation {
```
sofiaritz, line 56: Most of the checks here can be replaced with early returns, which would make the code more readable. :)

src/hosted_db/sqlite.rs
```diff
@@ -36,3 +47,35 @@ pub fn run_sqlite_query(path: &PathBuf, query: &str) -> Result<QueryResult, AybE
...
+            } else if result.stdout.len() > 0 {
+                let query_result: QueryResult = serde_json::from_str(&result.stdout)?;
+                Ok(query_result)
+            } else {
+                Err(AybError::Other {
+                    message: "No results from sandboxed query runner".to_string(),
+                })
+            }
+        }
+        None => Ok(query_sqlite(path, query)?.into()),
```
sofiaritz, line 79: You could remove a level of nesting by doing the following at the top
```rust
if isolation.is_none() {
   return Ok(query_sqlite(path, query)?.into());
}
```

sofiaritz (changes requested):
I like how you ended up implementing this, a clean and simple implementation :)
Now that the architecture of ayb is getting a little bit more complex, maybe it's time to create something like an [ARCHITECTURE.md](https://github.com/rust-lang/rust-analyzer/blob/d7c99931d05e3723d878bea5dc26766791fa4e69/docs/dev/architecture.md) file.

Cargo.toml
marcua: woohoo! learned something new! :)

scripts/build_nsjail.sh
marcua: Done

src/hosted_db/sandbox.rs
marcua: Done

src/bin/ayb_isolated_runner.rs
marcua: Done

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
+/* Retrieved and modified from
+  https://raw.githubusercontent.com/Defelo/sandkasten/83f629175d02ebc70fbb16b8b9e05663ea67ccc7/src/sandbox.rs
+  On December 6, 2023.
+  Original license: MIT.
```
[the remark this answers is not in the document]
marcua: Good call, done! :)

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+pub async fn run_in_sandbox(
+    nsjail: &Path,
+    db_path: &PathBuf,
+    query: &str,
+) -> Result<RunResult, AybError> {
+    let mut cmd = tokio::process::Command::new(nsjail);
+
+    cmd.arg("--really_quiet") // log fatal messages only
```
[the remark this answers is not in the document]
marcua: Interestingly, initially I tried importing Sandkasten so I could use its abstraction, but that broke down as that library wasn't designed as an abstraction for other libraries. So I copied in the parts that I needed here (thus the copyright notice), and realized that the amount of abstraction was too high for something `ayb` would have one concrete use case for. The abstracted code was more difficult to follow, since `ayb` isn't `nsjail-wrapper`, and so I went with a more concrete single use case here.

I generally avoid abstractions that have only one concrete instance since I can't vet the abstraction without multiple examples. My bias is to follow that general avoidance in this case, but if there's some specific benefit you can think of that outweighs the cost of indirection, I'm open to hearing it! :)

src/hosted_db/sandbox.rs
marcua: Done!

src/hosted_db/sandbox.rs
```diff
@@ -0,0 +1,120 @@
...
+        .arg("--iface_no_lo")
+        .args(["--mode", "o"]) // run once
+        .args(["--hostname", "ayb"])
+        .args(["--bindmount_ro", "/lib:/lib"])
+        .args(["--bindmount_ro", "/lib64:/lib64"])
+        .args(["--bindmount_ro", "/usr:/usr"])
+        .args(["--mount", "none:/tmp:tmpfs:size=100000000"]) // TODO(marcua): Restrict disk size more configurably?
+        // TODO(marcua): Set resource limits more configurably?
+        .args(["--max_cpus", "1"])
+        .args(["--rlimit_as", "64"]) // in MB
```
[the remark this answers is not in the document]
marcua: Perfect! Done!

src/hosted_db/sandbox.rs
marcua: Done!

src/hosted_db/sandbox.rs
marcua: Good eye! :) I was able to remove it with no effect.

src/hosted_db/sandbox.rs
marcua: This was originally a limit option to the Sandkasten implementation, and I figured providing a generous default was reasonable (the query would time out at this size of return anyway). That said, I was able to remove the `take` and make the readers `mut` and it seems to work just fine. We can always make this adjustable later. Good push to simplify this! :)

Cargo.toml
marcua: Done

src/http/structs.rs
marcua: Good push --- it was more laziness for not wanting to update GitHub actions to build `nsjail` :P.

We now have
* just the script
* GH actions builds `nsjail`
* documentation in README.md on isolation (yay!) and how to run tests (because you now have to build `nsjail` once, yay!)

Cargo.toml
marcua: Building `nsjail` solved it!!! :)

src/hosted_db/sqlite.rs
marcua: Nice! Thank you!

(I noticed `cargo clippy` also caught this, and realized clippy isn't breaking the build, so I created #250)

src/hosted_db/sqlite.rs
marcua: Done! Thank you!

src/hosted_db/sqlite.rs
marcua: Done!

src/hosted_db/sqlite.rs
marcua: Thank you for this feedback. Aside from the `match`, which we'll discuss below, are there others to consider?

src/hosted_db/sqlite.rs
marcua: I'm conflicted on this one. We'd get one level of nesting in a relatively short function, but in exchange
* I would stop using a language feature like `match`, which I feel is being used appropriately
* I'd have to introduce an `unwrap`, since the compiler won't be able to reason about the knowledge we'd gain with the `if`

Given we're using match in the way it was intended and forcing ourselves to enumerate all of the variants, I wonder if we'd gain much in exchange for reducing indentation in a short function.

src/http/config.rs
```diff
@@ -33,6 +33,7 @@ pub fn default_server_config() -> AybConfig {
             origin: "*".to_string(),
         },
         web: None,
+        isolation: None,
```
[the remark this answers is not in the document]
marcua: Very nice! Done!

marcua:
Thank you for the very helpful review @sofiaritz! You made the implementation an documentation far more approachable, and I appreciate your feedback. Looking forward to your next round!

Re: ARCHITECTURE.md, great idea. I added basic documentation on isolation and testing, but also created #252 for a deeper pass.

src/hosted_db/sandbox.rs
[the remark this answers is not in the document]
sofiaritz: You're right, that abstraction wouldn't make much sense in our case.
I think the comments are enough, I'll mark this as resolved :)

src/hosted_db/sqlite.rs
sofiaritz: This is just a style preference, the `Option` enum won't add more variants and thus `match` doesn't introduce any advantages over other ways of error checking/handling.

I feel that you are _really really_ skeptical about using `unwrap`/`expect`. Let's not forget that in our case when something panics inside a request to the server, the ayb server keeps running. In our case, using `unwrap`/`expect` in a way that can _actually_ panic isn't as bad as using it in a library or a TUI (not saying that that's the case with this suggestion, this is just a general comment).

I like early returns because they remove mental overhead from the developer and clearly separate the "invalid cases"/error handling from the actual logic of the code.

src/hosted_db/sqlite.rs
sofiaritz: Nope :)

sofiaritz:
Done! Just a few comments about early returns and `match`. Really excited for this to land! :)

src/hosted_db/sqlite.rs
sofiaritz: New idea! What about using `if let Some(isolation) = isolation` and returning the `None` case at the end of the function? Seems like a reasonable middleground :)

```rust
pub async fn potentially_isolated_sqlite_query(
    path: &PathBuf,
    query: &str,
    isolation: &Option<AybConfigIsolation>,
) -> Result<QueryResult, AybError> {
    if let Some(isolation) = isolation {
        let result = run_in_sandbox(Path::new(&isolation.nsjail_path), path, query).await?;
        if !result.stderr.is_empty() {
            let error: AybError = serde_json::from_str(&result.stderr)?;
            return Err(error);
        }
        
        if result.status != 0 {
            return Err(AybError::Other {
                message: format!(
                    "Error status from sandboxed query runner: {}",
                    result.status
                ),
            });
        }
        
        if !result.stdout.is_empty() {
            let query_result: QueryResult = serde_json::from_str(&result.stdout)?;
            return Ok(query_result);
        }
        
        return Err(AybError::Other {
            message: "No results from sandboxed query runner".to_string(),
        });
    }
    
    Ok(query_sqlite(path, query)?)
}
```

src/hosted_db/sandbox.rs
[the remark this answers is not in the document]
marcua: Hooray! Thank you!

src/hosted_db/sqlite.rs
marcua: Clean! I like it! I appreciated both of your bullet points on the broader point. I'll implement this and we should be good to go! Woohoo! Thank you!!!

sofiaritz (approved):
Amazing work!! When this lands, I'll update aybDocker and start hosting a public instance :)

marcua:
Thank you for the very helpful review, @sofiaritz! I'll merge now, and after #237 lands, I'll release the next version for your public instance! Woohooo! :)

Issue #237, opened 2023-12-15
User profiles


sofiaritz:
Add support for setting and getting user profiles with the following information:
* Display name (35 graphemes)
* Short description (160 chars -- that's what GitHub does, seems reasonable)
    * Maybe long description? A Markdown description with a character limit of something like 1000 chars, something like GitHub's profile READMEs
* Links (up to 4)
    * Support for verified links with rel=me
    * Maybe categories (website, work, social...) that can be used by UIs to display an icon?

Both the profile and each individual field can be unset.
* I'll probably submit a PR making the `db_interfaces` impl return an Option when needed instead of an error before submitting the user profiles PR

Feel free to assign this to me, I'm actively working on this :)

marcua:
Thank you for submitting this @sofiaritz ! You're now assigned and on the roadmap! :)

Display name, short description, and links sound great. Maybe see how it looks with or without a markdown description and we can decide.

Some thoughts on links
* on the frontend you might restrict up to 4 links, but on the backend we can model it as a list of links of any length so we can modify/iterate on the specific length later
* I'm observing that GitHub also allows you to reference an organization you work for, a location, and a website distinct from your social links.

I'm excited for this! It will allow ayb to be way more social! Thank you for working on it!

marcua:
Webfinger sounds like a cool addition! :)

To keep this issue focused on profiles, I'd be open to us pursuing it in a separate issue and we can discuss some of the details there. One thing we can consider in that issue is how to manage an increasing (N=2) number of URL templates (for the registration link and the profile page link). To keep the ayb.toml small and tidy, my thought would be to have these all in one place in the ayb.toml, or even served by the aybWeb server at a `.well-known`-like URL, and you can implicitly determine what functionality is enabled based on which URL templates are provided. We can discuss on that other issue, and I'm excited for webfinger! :)

sofiaritz:
> in a separate issue and we can discuss some of the details there

That's the idea, I'll open an issue :)

> To keep the ayb.toml small and tidy, my thought would be to have these all in one place

I think that would be confusing for new instance owners, also I don't think that a big `ayb.toml` file is necessarily a problem (eg. [Forgejo config cheatsheet](https://forgejo.org/docs/latest/admin/config-cheat-sheet/)). Instance admins should have the possibility of having a small ayb.toml with some features disabled and sane defaults, but also have to possibility of customizing most of the things in their instance (webfinger config, character limits, database limits, authentication config, etc).

[...]

PR #238, opened 2023-12-16
Improve database interfaces

sofiaritz:
This PR updates some functions from `db_interfaces` to return an Option instead of an error when `query_as` returns a `RowNotFound` error.

Related to #237

### DIFF
--- a/src/http/endpoints/create_database.rs
+++ b/src/http/endpoints/create_database.rs
@@ -24,7 +24,7 @@ async fn create_database(
     if entity.is_none() {
         return Err(AybError {
             message: format!("Entity not found: {:?}", entity_slug),
-        })
+        });
     }
 
     let entity = entity.unwrap();
--- a/src/http/endpoints/create_database.rs
+++ b/src/http/endpoints/create_database.rs
@@ -20,6 +20,14 @@ async fn create_database(
     let entity_slug = &path.entity;
     let entity = ayb_db.get_entity_by_slug(entity_slug).await?;
     let db_type = get_header(&req, "db-type")?;
+
+    if entity.is_none() {
+        return Err(AybError {
+            message: format!("Entity not found: {:?}", entity_slug),
+        })
+    }
+
+    let entity = entity.unwrap();
     let database = Database {
         entity_id: entity.id,
         slug: path.database.clone(),
--- a/src/ayb_db/db_interfaces.rs
+++ b/src/ayb_db/db_interfaces.rs
@@ -31,7 +31,7 @@ pub trait AybDb: DynClone + Send + Sync {
     ) -> Result<InstantiatedAuthenticationMethod, AybError>;
     async fn create_database(&self, database: &Database) -> Result<InstantiatedDatabase, AybError>;
     async fn get_or_create_entity(&self, entity: &Entity) -> Result<InstantiatedEntity, AybError>;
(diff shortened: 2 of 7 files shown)


src/http/endpoints/create_database.rs
```diff
@@ -20,6 +20,14 @@ async fn create_database(
     let entity_slug = &path.entity;
     let entity = ayb_db.get_entity_by_slug(entity_slug).await?;
     let db_type = get_header(&req, "db-type")?;
+
+    if entity.is_none() {
```
marcua, line 24: I'm trying to understand the benefit of this change toward `Option`. Originally the code returned as an `AybError`, which felt appropriate because you were requesting a row and that row didn't exist. This is similar to other ORMs (e.g., [Django](https://docs.djangoproject.com/en/5.0/topics/db/queries/#retrieving-a-single-object-with-get)). This refactor instead introduces `Option`, but then all callers then return their own `AybError`, which would have happened in the original implementation with less boilerplate.

Just trying to weigh the benefits and costs of the refactor. Thank you for being patient with me! :)

sofiaritz: A caller might need to handle null values themselves (eg. user profiles may be null) and thus it cannot be mixed with _real_ errors like a TLS failure or a timeout.
Also, related to #103, different endpoints calling the same db function may need to return their own error code/error message instead of one determined by the database interface.

This PR is just a clean-up for future features (better errors, user profiles, etc).
I hope this explains the purpose if the PR :)

marcua: I totally agree the caller *might* want to capture the case where a row doesn't exist. To your point, I think the solution is #103. Here's why I think we should consider moving to AybError with specific subtypes like RecordNotFound:
* It allows you to capture the error and handle it if you want to (you can choose to see if the return type is an `Err(RecordNotFound...)`)
* It allows you to not capture the error if you want to (meaning we wouldn't need the additional code in the call sites that this PR introduces), since you're not forced to pick between Some and None
* There's precedent in other ORMs (like Django, which is the only one I'm closely familiar with to be honest) to consider it an error
* It keeps the `list_...` and `get_...` behaviors more consistent in their return types, though only sort of (this is a weak argument)
* It moves us closer to #103, which is broadly good for the project (you've said this, so I'm not telling you anything new with this :))

I wonder if we can start toward #103 without requiring an codebase-wide rewrite. If we made AybError have just two variants for now:
```rust
enum AybError {
    RecordNotFound,
    DeprecatedAybError { message: String },
}
```

We could find and replace al instances of `AybError` construction with `AybError::DeprecatedAybError` construction for now, and only add the special variant as part of this PR. As a result, we could add all the special variants for other use cases in another PR while keeping this one small.

I know it's frustrating to have so much scrutiny on a small PR that is addressing a real issue, so I hope you'll have patience with me as we sort this out :).

sofiaritz: I think you're mostly right, I'll close this PR and implement the _improved_ AybError instead :)
That way I can implement the same idea as this PR and also partially fix #103. (Special offer! 2✕1 :p)

marcua: Amazing! Thank you so much for doing this patiently! :)

sofiaritz: No problem! :)

PR #240, opened 2023-12-17
`TemplateString` struct

sofiaritz:
This struct can be used in `AybConfig` to define a template string, it is both serialized and deserialized as a string. This _wrapper_ struct provides the `execute` function, which provided a Vec of string tuples replaces the key (left value) with the value (right value).

Related to a conversation in #237, shouldn't conflict with #238

### DIFF
--- a/src/email/templating.rs
+++ b/src/email/templating.rs
@@ -1,19 +1,25 @@
 use crate::http::structs::AybConfigEmail;
-
-const CLI_CONFIRM_TMPL: &str = "To complete your registration, type\n\tayb client confirm {token}";
-const WEB_CONFIRM_TMPL: &str = "To complete your registration, visit\n\t {url}";
+use crate::templating::TemplateString;
 
 pub fn render_confirmation_template(config: &AybConfigEmail, token: &str) -> String {
+    let cli_confirm_tmpl: TemplateString =
+        "To complete your registration, type\n\tayb client confirm {token}"
+            .to_string()
+            .into();
+    let web_confirm_tmpl: TemplateString = "To complete your registration, visit\n\t {url}"
+        .to_string()
+        .into();
+
     if let Some(tmpl_conf) = &config.templates {
         if let Some(confirm_conf) = &tmpl_conf.confirm {
-            return WEB_CONFIRM_TMPL.replace(
-                "{url}",
+            return web_confirm_tmpl.execute(vec![(
+                "url",
                 &confirm_conf
                     .confirmation_url
-                    .replace("{token}", &urlencoding::encode(token)),
-            );
+                    .execute(vec![("token", &urlencoding::encode(token))]),
+            )]);
         }
     }
 
-    CLI_CONFIRM_TMPL.replace("{token}", token)
+    cli_confirm_tmpl.execute(vec![("token", &urlencoding::encode(token))])
 }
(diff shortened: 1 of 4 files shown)


src/templating.rs
```diff
@@ -0,0 +1,66 @@
...
+pub struct TemplateString {
+    string: String,
+}
+
+impl TemplateString {
+    pub fn execute(&self, values: Vec<(&str, &str)>) -> String {
+        let mut string = self.string.clone();
+        values
+            .iter()
+            .for_each(|(k, v)| string = string.replace(&format!("{{{}}}", k), v));
```
sofiaritz, line 15: The double brackets are used to escape them:
```
assert_eq!(format!("{{"), "{".to_string())
```

src/templating.rs
```diff
@@ -0,0 +1,66 @@
+use serde::de::{Error, Visitor};
```
marcua, line 1: Unlike my broader bias toward end-to-end tests, this one feels ripe for a few tiny unit tests in the module itself :)

src/email/templating.rs
```diff
@@ -1,19 +1,25 @@
 use crate::http::structs::AybConfigEmail;
-
-const CLI_CONFIRM_TMPL: &str = "To complete your registration, type\n\tayb client confirm {token}";
-const WEB_CONFIRM_TMPL: &str = "To complete your registration, visit\n\t {url}";
+use crate::templating::TemplateString;
 
 pub fn render_confirmation_template(config: &AybConfigEmail, token: &str) -> String {
+    let cli_confirm_tmpl: TemplateString =
```
marcua, line 5: Would it be possible to still keep them module-level CLI_CONFIRM_TMPL so we don't have to compile them on every render?

marcua (changes requested):
This one is great, just two tiny suggestions. Thank you for submitting it! :)

src/email/templating.rs
sofiaritz: I'm not sure whether I could make the struct const... I'll try, in either case this should be heavily optimized by the compiler

src/templating.rs
sofiaritz: Haha will do :)

src/email/templating.rs
sofiaritz: Looks like I can't make it const without polluting the code with lifetimes everywhere. This shouldn't have a huge performance overhead anyway :)

src/templating.rs
sofiaritz: Done! :)

src/email/templating.rs
```diff
@@ -1,19 +1,25 @@
...
                     .confirmation_url
-                    .replace("{token}", &urlencoding::encode(token)),
-            );
+                    .execute(vec![("token", &urlencoding::encode(token))]),
+            )]);
         }
     }
 
-    CLI_CONFIRM_TMPL.replace("{token}", token)
+    cli_confirm_tmpl.execute(vec![("token", &urlencoding::encode(token))])
```
marcua, line 24: ah whoops sorry I didn't catch it sooner --- do we need the CLI token to be URL-encoded?

marcua (changes requested):
Great tests and all good on not keeping it in global scope (probably a good thing :)). One tiny thing I missed in the previous review and we should be good to go! :)

src/email/templating.rs
sofiaritz: Ooops, we don't, I'll change that :p

sofiaritz:
Round three! :) :woman_cartwheeling:

marcua (approved):
Thank you for getting this just right! :)

PR #241, opened 2023-12-17
Replace `AybError` with an enum

sofiaritz:
In this initial iteration only two variants are defined: `RecordNotFound` and `Other`, the second one contains a message.

The enum is tagged under "type". This means that a JSON response would look like this:
```json
{
    "type": "Other",
    "message": "..."
}
```

Related to #103 and #238.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,5 @@
 use actix_web;
-use derive_more::{Display, Error};
+use derive_more::Error;
 use fernet;
 use lettre;
 use prefixed_api_key;
@@ -9,14 +9,21 @@ use rusqlite;
 use serde::{Deserialize, Serialize};
 use serde_json;
 use sqlx;
+use std::fmt::{Display, Formatter};
 use std::string;
 use toml;
 
-#[derive(Debug, Deserialize, Display, Error, Serialize)]
+#[derive(Debug, Deserialize, Error, Serialize)]
 #[serde(tag = "type", content = "value")]
 pub enum AybError {
-    RecordNotFound,
+    RecordNotFound { id: String, record_type: String },
     Other { message: String },
+}
+
+impl Display for AybError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
 }
 
 impl actix_web::error::ResponseError for AybError {
--- a/src/error.rs
+++ b/src/error.rs
@@ -115,8 +115,11 @@ impl From<std::io::Error> for AybError {
 
 impl From<sqlx::Error> for AybError {
     fn from(cause: sqlx::Error) -> Self {
-        AybError::Other {
-            message: format!("{:?}", cause),
+        match cause {
+            sqlx::Error::RowNotFound => Self::RecordNotFound,
+            _ => Self::Other {
+                message: format!("{:?}", cause),
+            }
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -115,11 +115,8 @@ impl From<std::io::Error> for AybError {
 
 impl From<sqlx::Error> for AybError {
     fn from(cause: sqlx::Error) -> Self {
-        match cause {
-            sqlx::Error::RowNotFound => Self::RecordNotFound,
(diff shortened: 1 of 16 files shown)


sofiaritz:
Oh wait, I just found something that can be improved, give me a second

sofiaritz:
Done! :)

src/error.rs
```diff
@@ -25,119 +27,122 @@ impl actix_web::error::ResponseError for AybError {
...
         }
     }
 }
 
 impl From<sqlx::Error> for AybError {
     fn from(cause: sqlx::Error) -> Self {
-        AybError {
-            message: format!("{:?}", cause),
+        match cause {
+            sqlx::Error::RowNotFound => Self::RecordNotFound,
```
marcua, line 119: Could we catch it closer to the original location so we can include a message like "Entity not found" with an ID/slug?

marcua (changes requested):
Quick question for you --- sorry if my pseudocode misled you --- you could imagine including a type of record and string identifier to help with debugging :)

src/error.rs
sofiaritz: > Could we catch it closer to the original location so we can include a message like "Entity not found" with an ID/slug?

Yes, of course :)

> You could imagine including a type of record and string identifier to help with debugging :)

I'm not really sure I follow, something like this?
```rust
enum AybError {
    RecordNotFound { id: String, record_type: String },
    ...
}
```

marcua: > I'm not really sure I follow, something like this?
```rust
enum AybError {
    RecordNotFound { id: String, record_type: String },
    ...
}

Exactly like that! :) I think it's OK that `id` is a `String` even if it's sometimes an int and sometimes a String, since it's just for helping figure out what lookup went wrong.

**Review by marcua — CHANGES_REQUESTED** [review]
Great job on this. Your proposal sounds great.

Also just checking: tests pass as-is?

**Comment by sofiaritz** [comment]
They did, now they don't :p
Looks like I have to update the E2E tests _or_ the client implementation to display the message instead of the raw error:
https://github.com/sofiaritz/ayb/actions/runs/7241116339/job/19724857624

What do you think? :)

File: src/error.rs [hunk 2]
```diff
@@ -9,16 +9,23 @@ use rusqlite;
 use serde::{Deserialize, Serialize};
 use serde_json;
 use sqlx;
+use std::fmt::{Display, Formatter};
 use std::string;
 use toml;
 
-#[derive(Debug, Deserialize, Display, Error, Serialize)]
+#[derive(Debug, Deserialize, Error, Serialize)]
 #[serde(tag = "type", content = "value")]
```
marcua, line 17: Do we think it's important to have content=value? I like tagging the type, but do we think it helps to have an extra layer of nesting (and bulking up of the serialized representation? (I'm OK skipping it.)

marcua:
Oh joy! :)

I'd either update the client implementation to display just the message for other, or add the logic to the `Display`/`fmt` function (we can eventually have different formatting for different variants).

src/error.rs
sofiaritz: I think it is, at least for UI implementors.

sofiaritz:
I'll add the logic to the `Display` trait :)

src/error.rs
marcua: The type of the response will be from tag=type, no? I'm curious about content=value, which adds a level of nesting but as best I can tell doesn't provide additional information. It's possible I'm wrong about that, though! :)

marcua (changes requested):
One tiny additional detail about content=value :)

src/error.rs
sofiaritz: Yes yes, you're completely right. I realized the example I had in mind doesn't make any sense :sweat_smile:  :p
I'll change it :)

Sorry :p

sofiaritz:
Done, I had a mental oopsies yesterday while thinking about content=value :p

Ready for review again :)
(Also, tests pass https://github.com/sofiaritz/ayb/actions/runs/7260945881/job/19781142417)

marcua:
Thank you so much as always @sofiaritz ! Looks great, merging! :)

Issue #239, opened 2023-12-17
Webfinger support


sofiaritz:
Right now UIs have no way to link users on other instances without some _magic guessing_. Implementing [Webfinger](https://webfinger.net) would allow users to mention users on any Webfinger-compliant service (ayb, Mastodon, Akkoma, etc.) and link to its profile page.

In the future, this could also be used to link to the entity retrieval endpoint with the addition of a link of type `application/aybEntity+json`.

marcua:
This is great! Before we start it, there are two things to consider

## The use case
While in theory adding webfinger support sounds good, could you describe the user experience a little more? Specifically I'm wondering where and how ayb users might mention other services, and where users on other services might mention ayb users?

## How to configure it
I did a bad job of describing it in #237, but while I'm not opposed to adding this stuff to configuration files, I'm trying to think of a way to standardize it. I can think of two approaches

### One section of `ayb.toml`
We could have a `web` section of `ayb.toml` where all configuration will go. For example

```toml
[web]
server_url = "https://aybWeb.sofiaritz.com/some/prefix/"
profile_path = "u/{entity}"
registration_confirm_path = "register/confirm/{token}"
```
The benefit of this approach is all of the web frontend related stuff is in one section, rather than having URLs in several sections.
(Note that we'd have to change how registration confirmation works a little bit, since you'd only have a registration URL and not other content, which the email logic would have to handle. That seems fine for standardization.)

### Have the web frontend define its paths
The `ayb` server shouldn't be responsible for knowing specific paths on a remote server, and the server is better-suited for providing its paths.  So ayb could have a simpler configuration like

```toml
[web]
server_url = "https://aybWeb.sofiaritz.com/"
```

And the aybWeb server could provide a file with its routes/paths. This could be in an [OpenAPI](https://swagger.io/specification/) spec or a way simpler format. For example:
https://aybWeb.sofiaritz.com/.well-known/ayb.json
```json`
{
  paths: {
    profile_path: "/some/prefix/u/{entity}"
    registration_confirm_path: "/some/prefix/register/confirm/{token}"
  }
}
```

The benefit of this approach is that the responsibility lies with the web frontend for communicating its paths, which it is better suited to doing. I think it adds a bit of complexity at this stage though, and requires some bootstrapping logic web an `ayb` server first starts to query the web frontend, so perhaps approach #1 is better in its simplicity.

sofiaritz:
> could you describe the user experience a little more

The process is the following:
[...]

PR #242, opened 2023-12-19
Implement webInfo

sofiaritz:
webInfo allows ayb to retrieve endpoints from a given frontend.

This commits removes the `email.templates` object from `ayb.toml` and adds `web.info_url` of type `Url`.

The webInfo endpoint (`web.info_url`) is queried on server startup. An example of the response is provided at `docs/config/endpoints/example.json` and the schema is also provided at `docs/config/endpoints/schema.json`.

Related to #239


docs/config/endpoints/example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://raw.githubusercontent.com/marcua/ayb/main/docs/config/endpoints/schema.json",
+  "base_url": "https://aybWeb.sofiaritz.com",
+  "endpoints": {
+    "profile": "/u/{entity}",
+    "confirmation": "/auth/confirm/{token}"
+  }
+}
```
marcua, line 8:
```suggestion
}

```

docs/config/endpoints/schema.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://raw.githubusercontent.com/marcua/ayb/main/docs/config/endpoints/schema.json",
+  "title": "ayb web information",
+  "description": "Information used by the ayb server to interact with the UI",
```
marcua, line 5:
```suggestion
  "description": "Information used by the ayb server to interact with a web-based UI",
```

docs/config/endpoints/schema.json
```diff
@@ -0,0 +1,32 @@
...
+      "description": "The base URL",
+      "examples": ["https://aybWeb.sofiaritz.com/"],
+      "type": "string"
+    },
+    "endpoints": {
+      "description": "The endpoints of the frontend",
+      "type": "object",
+      "properties": {
+        "profile": {
+          "description": "The profile page",
```
marcua, line 18:
```suggestion
          "description": "The profile page for an entity",
```

docs/config/endpoints/schema.json
```diff
@@ -0,0 +1,32 @@
...
+          "examples": ["/u/{entity}"],
+          "type": "string"
+        },
+        "confirmation": {
+          "description": "The confirmation endpoint",
```
marcua, line 23:
```suggestion
          "description": "The account confirmation endpoint",
```

docs/config/endpoints/schema.json
```diff
@@ -0,0 +1,32 @@
...
+          "examples": ["/auth/confirm/{entity}"],
+          "type": "string"
+        }
+      },
+      "required": ["profile", "confirmation"]
+    }
+  },
+  "required": ["endpoints"]
+}
```
marcua, line 32:
```suggestion
}

```

docs/config/endpoints/schema.json
```diff
@@ -0,0 +1,32 @@
...
+        "confirmation": {
+          "description": "The confirmation endpoint",
+          "examples": ["/auth/confirm/{entity}"],
+          "type": "string"
+        }
+      },
+      "required": ["profile", "confirmation"]
+    }
+  },
+  "required": ["endpoints"]
```
marcua, line 31: is `base_url` required? (I think yes)

src/http/config.rs
```diff
@@ -28,11 +28,11 @@ pub fn default_server_config() -> AybConfig {
             smtp_port: 465,
             smtp_username: "login@example.org".to_string(),
             smtp_password: "the_password".to_string(),
-            templates: None,
         },
         cors: AybConfigCors {
             origin: "*".to_string(),
         },
+        web: None,
```
marcua, line 35: is there an easy way to add coverage to the end-to-end test around this? low bar, let's create an issue to cover the web information functionality in end-to-end tests in the future.

src/http/server.rs
```diff
@@ -80,6 +81,15 @@ pub async fn run_server(config_path: &PathBuf) -> std::io::Result<()> {
     let ayb_conf_for_server = ayb_conf.clone();
     fs::create_dir_all(&ayb_conf.data_path).expect("Unable to create data directory");
     let ayb_db = connect_to_ayb_db(ayb_conf.database_url).await.unwrap();
+    let web_info = if let Some(web_conf) = ayb_conf.web {
+        Some(
+            WebInfo::from_url(&web_conf.info_url)
+                .await
+                .expect("failed to retrieve information from the web frontend"),
```
marcua, line 88:
```suggestion
                .expect("Failed to retrieve information from the web frontend"),
```

src/web_info.rs
```diff
@@ -0,0 +1,43 @@
+use crate::error::AybError;
```
marcua, line 1: A few thoughts on naming and organization
* We should consider this living in `src/http/`: it might not look like it now, but in the future, ayb might support other protocols (e.g., psql)
* `WebInfo` -> `WebFrontendDetails` or `WebFrontendConfig`? Trying to be more specific about what web information it is.

src/web_info.rs
```diff
@@ -0,0 +1,43 @@
...
+    endpoints: WebEndpoints,
+}
+
+#[derive(Clone, Deserialize)]
+pub struct WebEndpoints {
+    profile: TemplateString,
+    confirmation: TemplateString,
+}
+
+impl WebInfo {
```
marcua, line 18: I found this implementation to be quite elegant and compact! :)

src/web_info.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub async fn from_url(url: &Url) -> Result<Self, AybError> {
+        Ok(reqwest::get(url.to_string()).await?.json().await?)
+    }
+
+    pub fn profile(&self, entity: &str) -> String {
+        let relative = self.endpoints.profile.execute(vec![("entity", entity)]);
+        let absolute = self
+            .base_url
+            .join(&relative)
+            .expect("invalid profile template string provided by the web frontend");
```
marcua, line 28:
```suggestion
            .expect("Invalid profile template string provided by the web frontend");
```

src/web_info.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    pub fn confirmation(&self, token: &str) -> String {
+        let relative = self
+            .endpoints
+            .confirmation
+            .execute(vec![("token", &urlencoding::encode(token))]);
+        let absolute = self
+            .base_url
+            .join(&relative)
+            .expect("invalid confirmation template string provided by the web frontend");
```
marcua, line 40:
```suggestion
            .expect("Invalid confirmation template string provided by the web frontend");
```

src/web_info.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[derive(Clone, Deserialize)]
+pub struct WebEndpoints {
+    profile: TemplateString,
+    confirmation: TemplateString,
+}
+
+impl WebInfo {
+    pub async fn from_url(url: &Url) -> Result<Self, AybError> {
+        Ok(reqwest::get(url.to_string()).await?.json().await?)
```
marcua, line 20: Given the templates evaluate the same way each time, can we push the logic of running `profile()`/`confirmation()` into the constructor, make the struct fields `pub`, and make the methods private? That way we don't have to execute a template and compute a URL each time we need it.

marcua (changes requested):
Thank you for another great implementation! Some suggestions/comments over to you! :)

src/http/server.rs
sofiaritz: Expect messages should be undercased, that's how std and most libraries do it
Example: https://doc.rust-lang.org/std/option/enum.Option.html#method.expect

src/web_info.rs
sofiaritz: See previous comment

src/web_info.rs
sofiaritz: See previous comment

src/web_info.rs
sofiaritz: `WebFrontendDetails` seems more accurate, I'll move the module and update its name

src/http/config.rs
sofiaritz: I checked if reqwest had support for data URIs, but that's not the case, so there isn't an easy way to E2E test this.
I can add the E2E tests in a follow-up PR, unless you want both things to land together :)

docs/config/endpoints/schema.json
sofiaritz: Yes it is, I forgot to add it there :p

src/web_info.rs
sofiaritz: Thank youu! :)

sofiaritz:
Ready for the second round! :)
Just a comment about `expect` messages and E2E

src/http/server.rs
marcua: I learn something new every day! Thank you! :)

src/http/config.rs
marcua: I think creating an issue and doing it in a follow-up PR is just fine! Thank you! :)

src/web_info.rs
marcua: Thank you! :)

marcua (approved):
Another one done! Thank you @sofiaritz! :)

PR #248, opened 2023-12-24
Implement user profiles and rel=me verification

### LINKED ISSUE #237: User profiles

sofiaritz:
Christmas came early this year :)

Closes #237


Cargo.toml
```diff
@@ -22,7 +22,7 @@ reqwest = { version = "0.11.22", features = ["json"] }
 rusqlite = { version = "0.27.0", features = ["bundled"] }
 regex = { version = "1.10.2"}
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0.108" }
+serde_json = { version = "1.0.108", features = ["raw_value"] }
```
sofiaritz, line 25: This is needed for `sqlx::types::Json`

Cargo.toml
```diff
@@ -32,6 +32,7 @@ prettytable-rs = { version = "0.10.0"}
 urlencoding = { version = "2.1.3" }
 actix-cors = { version = "0.6.5" }
 url = { version = "2.5.0", features = ["serde"] }
+html_parser = { version = "0.7.0" }
```
marcua, line 35: I know very little about Rust HTML5 parsers. Is this the right one for the project vs. `html5ever`, etc.? (Really, no informed opinion here, just curiosity.)

migrations/postgres/20231223165738_user_profiles.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE entity ADD display_name VARCHAR(35);
```
marcua, line 1: Whereas SQLite doesn't support adding multiple columns in a single `ALTER`, I believe Postgres does.

src/http/verification.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::str::FromStr;
+use url::Url;
+
+pub async fn is_verified(input_url: Url, expected_url: Url) -> bool {
+    let website = reqwest::get(input_url.to_string()).await;
+    if website.is_err() {
+        return false;
+    }
+
+    let website = website.unwrap().text().await;
```
marcua, line 11: Curiously, is there a Rusty way to avoid the unwraps? I know you're trying to avoid levels of nesting, and you're checking for is_err on previous lines, but I always get nervous at unwraps that might panic. If I should just be OK with it given Rust programming style, all good, but I figured I'd ask since I'm trying to avoid unwrapping in general and as a result do more nesting to convince myself I handle all cases.

src/http/verification.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    let website = Dom::parse(&website.unwrap());
+    if website.is_err() {
+        return false;
+    }
+
+    let website = website.unwrap();
+    for children in website.children {
+        if let Some(element) = children.element() {
```
marcua, line 23: Is there a way to provide an xpath (or other path query) to the "`a` with an `href` with a `rel=me`" so that you can push the search into the (presumably faster) library and avoid having to process every element of the website linearly?

src/http/structs.rs
```diff
@@ -100,9 +100,24 @@ pub struct EntityPath {
...
+pub struct EntityProfileLink {
+    pub url: String,
+    pub verified: bool,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EntityProfile {
+    pub display_name: Option<String>,
+    pub description: Option<String>,
+    pub workplace: Option<String>,
```
marcua, line 113: drastically overthinking it: do we want to more generically call it an `organization`, in case someone has affinity to it but doesn't identify as working there?

src/http/structs.rs
```diff
@@ -100,9 +100,24 @@ pub struct EntityPath {
...
 }
 
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EntityProfileLink {
+    pub url: String,
+    pub verified: bool,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EntityProfile {
```
marcua, line 110: is location in the world interesting for a profile?

src/http/structs.rs
```diff
@@ -100,9 +100,24 @@ pub struct EntityPath {
...
+    pub display_name: Option<String>,
+    pub description: Option<String>,
+    pub workplace: Option<String>,
+    pub links: Vec<EntityProfileLink>,
+}
+
 #[derive(Serialize, Deserialize, Clone)]
```
marcua, line 117: I don't have a fully formed opinion, but I'm curious if we want to have a separation between `Entity` and `EntityQueryResponse` as you have in this PR. Are we separating them because we believe you should have to explicitly request this metadata separate from stuff like the type of entity? Is that for performance reasons or usability reasons?

src/ayb_db/db_interfaces.rs
```diff
@@ -247,6 +260,36 @@ WHERE id = $1
...
+    display_name = $3,
+    links = $4
+WHERE entity.id = $5
+RETURNING id, slug, entity_type, display_name, description, workplace, links
+                    "#
+                )
+                .bind(&entity.description)
+                .bind(&entity.workplace)
+                .bind(&entity.display_name)
+                .bind(serde_json::to_value(&entity.links).unwrap())
```
marcua, line 278: safe unwrap?

src/http/endpoints/update_profile.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[derive(Deserialize)]
+struct ProfileUpdate {
+    pub display_name: Option<String>,
+    pub description: Option<String>,
+    pub workplace: Option<String>,
+    pub links: Vec<ProfileLinkUpdate>,
+}
+
+#[post("/v1/entity/{entity}")]
```
marcua, line 26: Not that we're RESTful anyway, but just thinking through POST/PUT/PATCH

src/http/verification.rs
```diff
@@ -0,0 +1,39 @@
+use html_parser::Dom;
```
marcua, line 1: Should the module be named url_verification?

src/http/verification.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::str::FromStr;
+use url::Url;
+
+pub async fn is_verified(input_url: Url, expected_url: Url) -> bool {
```
marcua, line 5: is_verified_url?

src/http/verification.rs
```diff
@@ -0,0 +1,39 @@
...
+    let website = reqwest::get(input_url.to_string()).await;
```
marcua, line 6: I believe that to be verified, the target has to be `https` as well

src/http/endpoints/update_profile.rs
```diff
@@ -0,0 +1,86 @@
...
+    let links = if let Some(web_info) = Option::as_ref(&**web_info) {
+        let mut links = vec![];
+        for link in profile.links.into_iter() {
+            let url = Url::parse(&link.url)?;
+            links.push(Link {
+                url: link.url,
+                verified: is_verified(
+                    url,
+                    Url::from_str(&web_info.profile(entity_slug))
+                        .expect("invalid web frontend url"),
```
marcua, line 52: Is there a way to avoid a panic here?

src/http/endpoints/update_profile.rs
```diff
@@ -0,0 +1,86 @@
...
+    let authenticated_entity = unwrap_authenticated_entity(&authenticated_entity)?;
+    let entity_slug = &path.entity;
+    let profile = profile.into_inner();
+
+    if entity_slug != &authenticated_entity.slug {
+        return Ok(HttpResponse::Unauthorized().finish());
+    }
+
+    let instantiated_entity = ayb_db.get_entity_by_slug(entity_slug).await?;
+    let links = if let Some(web_info) = Option::as_ref(&**web_info) {
```
marcua, line 43:
```suggestion
    let links = if let Some(web_info) = Option::as_ref(&**web_info) {
    // If there is a known web frontend, we verify the identity of links.
```

src/http/endpoints/update_profile.rs
```diff
@@ -0,0 +1,86 @@
...
+                    url,
+                    Url::from_str(&web_info.profile(entity_slug))
+                        .expect("invalid web frontend url"),
+                )
+                .await,
+            })
+        }
+
+        Some(links)
+    } else {
```
marcua, line 59:
```suggestion
    } else {
    // If there is no known web frontend, we save links to the profile without verification.
```

src/ayb_db/models.rs
```diff
@@ -166,13 +166,27 @@ pub struct InstantiatedDatabase {
 pub struct Entity {
     pub slug: String,
     pub entity_type: i16,
+    pub display_name: Option<String>,
+    pub description: Option<String>,
+    pub workplace: Option<String>,
+    pub links: Option<Vec<Link>>,
```
marcua, line 172: sanity-checking only the instantiated one should have the sqlx JSON wrapper

src/http/verification.rs
marcua, line 5: Docstring along the lines of "Verifies the identity of an HTTPS link URL if we find a `rel=me` link pointing to `expected_url` on the `input_url."

src/http/endpoints/update_profile.rs
```diff
@@ -0,0 +1,86 @@
...
+use crate::http::structs::EntityPath;
+use crate::http::utils::unwrap_authenticated_entity;
+use crate::http::verification::is_verified;
+use crate::http::web_frontend::WebFrontendDetails;
+use actix_web::{post, web, HttpResponse};
+use serde::Deserialize;
+use std::str::FromStr;
+use url::Url;
+
+#[derive(Deserialize)]
```
marcua, line 13: So nice to have this endpoint-specific struct stuff in the same file as the endpoint. I'm glad you cleaned up our endpoints earlier! :)

marcua (changes requested):
WOW! A present indeed! :)

Thank you so much for putting this together. It's so comprehensive and even adds identity verification! Amazing!

src/ayb_db/db_interfaces.rs
sofiaritz: > Safe unwrap?

Yes, this unwrap does not refer to the datatype, this unwrap can only panic if the struct being serialized contains a `HashMap` whose keys are not strings.
That's not the case, so it's safe to unwrap here :)
I don't know if AybError implements `From<serde_json::Error>`, if that's the case I'll replace the unwraps with the try operator (`?`)

I think the datatype is checked by sqlx

src/http/endpoints/update_profile.rs
sofiaritz: I think PATCH is more accurate, in the previous implementation POST made more sense, but now it doesn't :p

src/http/verification.rs
sofiaritz: I don't think that's the case, at least I couldn't find any documentation/wiki that said only HTTPS sites can be verified.

src/http/endpoints/update_profile.rs
sofiaritz: This panic is similar to this one: https://github.com/marcua/ayb/blob/7bcd087028b86e1dcce4b552263ff6a5ba186293/src/http/web_frontend.rs#L28

It only can happen when the web frontend didn't send the correct endpoints, which shouldn't happen ever. And if it does happen, the instance administrator must know that something is off.

From the Rust book:
> It’s advisable to have your code panic when it’s possible that your code could end up in a bad state. In this context, a bad state is when some assumption, guarantee, contract, or invariant has been broken, such as when invalid values, contradictory values, or missing values are passed to your code.

In that case, the web frontend has sent invalid values and broken the guarantee that all endpoints are valid URLs.

src/http/structs.rs
sofiaritz: That's the word I was looking for! I'll change it :)

src/http/verification.rs
sofiaritz: One of the cases where it's okay to unwrap/expect is when you have more information than the compiler.
From the Rust book:
> It would also be appropriate to call `unwrap` or `expect` when you have some other logic that ensures the `Result` will have an `Ok` value, but the logic isn’t something the compiler understands. You’ll still have a `Result` value that you need to handle: whatever operation you’re calling still has the possibility of failing in general, even though it’s logically impossible in your particular situation. 

We've checked that the above isn't an error, then that panic can't happen :)

src/http/structs.rs
sofiaritz: Yes, I'll add it :)

src/http/endpoints/update_profile.rs
sofiaritz: :)

Cargo.toml
sofiaritz: This is just my go-to parser, but after some research looks like using an `html5ever`-based parser is going to be better.
I'll use `scraper`, an `html5ever`-based crate with `selectors` :)

src/http/verification.rs
sofiaritz: I've found a clean way to avoid the unwraps :)

src/http/structs.rs
sofiaritz: `Entity` doesn't contain things like the databases owned by the entity, that's why it's separated :)

src/ayb_db/models.rs
sofiaritz: Exactly :)

sofiaritz:
Finally got to time to work on this :)
Done! Back to you :p

Cargo.toml
marcua: Wow! This ended up so clean! :)

src/http/structs.rs
marcua: Owned databases makes sense to me, I'm more curious about the fields directly on the Entity struct/table. I'm good either way, and I get the sense your bias is toward separating them, which is OK with me. Just wanted to sanity-check! :)

src/ayb_db/db_interfaces.rs
marcua: Perfect! Thank you!

src/http/endpoints/update_profile.rs
```diff
@@ -19,11 +19,12 @@ struct ProfileLinkUpdate {
...
     pub display_name: Option<String>,
     pub description: Option<String>,
-    pub workplace: Option<String>,
+    pub organization: Option<String>,
+    pub location: Option<String>,
     pub links: Vec<ProfileLinkUpdate>,
 }
 
-#[post("/v1/entity/{entity}")]
+#[patch("/v1/entity/{entity}")]
```
marcua, line 27: I was like "why does the client still work" and then I realized we haven't updated the client or covered the endpoint with end-to-end tests. Whoopsie! :)

src/http/verification.rs
marcua: I thought I read it somewhere, and found it in the [Mastodon docs on verification](https://docs.joinmastodon.org/user/profile/#verification). I suspect the reason is that if the link isn't HTTPS, you could be tricked into verifying through a man-/person-in-the-middle attack.

src/http/endpoints/update_profile.rs
marcua: Great explanation, thank you!

marcua (changes requested):
Great job on iterations! Some tiny responses and one realization about client/tests. Thank you so much for this work! :)

src/http/endpoints/update_profile.rs
sofiaritz: Hahaha, you're right. Should I create an ayb client command for getting and updating the profile? Or should I just send requests to the endpoint in the E2E tests? :)

src/http/verification.rs
sofiaritz: You're right, I'll update the implementation to return false on non-HTTPS URLs :)
Attackers won't have an good time with this :smiling_imp:

src/http/structs.rs
sofiaritz: Alright, I'll keep them separated (at least for now). You've made me realize that the entity retrieval endpoint doesn't send the entity type. I'll change that in a future PR :)

sofiaritz:
I've updated the URL verification module to not verify non-HTTPS URLs :)
Left a question about how should I approach the testing of this endpoint

src/http/endpoints/update_profile.rs
marcua: Great question! While it's a little clunky, an `ayb` client command would be great. Not to be too prescriptive (you can decide as you play with it), but something like

`ayb client update_profile sofiaritz links https://link.one https://link.two`
(links is the hardest, but `organization`, `location`, etc. would be similar but with only one arg in the list)

marcua: ...and I guess something similar for `get_profile?` (I remember a discussion when you introduced the details enpoint that you wanted to keep the database list and the other profile information separate so table/csv export would still work, but I don't feel strongly either way.)

sofiaritz:
Oh no... merge conflicts!!! :cry:

sofiaritz:
Done! Tests pass :) https://github.com/sofiaritz/ayb/actions/runs/7347777240/job/20004756637

I've introduced the `ayb client profile <entity>` and `ayb client update_profile <entity> --{meta_to_update} <value>`, both are E2E tested :)

src/bin/ayb.rs
```diff
@@ -129,6 +130,27 @@ async fn main() -> std::io::Result<()> {
...
+                )
+                .subcommand(
+                    Command::new("update_profile")
+                        .about("Update the profile of an entity")
+                        .arg(arg!(<entity> "The entity to update").required(true))
+                        .arg(arg!(--display_name <value> "New display name").required(false))
+                        .arg(arg!(--description <value> "New description").required(false))
+                        .arg(arg!(--organization <value> "New organization").required(false))
+                        .arg(arg!(--location <value> "New location").required(false))
+                        .arg(arg!(--link <value> "New URL, will overwrite previous ones").required(false))
```
marcua, line 152: given you can have multiple links, should it be possible to send a list with `.num_args(0..)`? we could call it `--links` at that point as well.

src/bin/ayb.rs
```diff
@@ -214,19 +236,78 @@ async fn main() -> std::io::Result<()> {
...
+                        Err(err) => println!("Error: {}", err),
+                    }
+                }
+            } else if let Some(matches) = matches.subcommand_matches("update_profile") {
+                if let Some(entity) = matches.get_one::<String>("entity") {
+                    let update_profile = {
+                        match client.entity_details(entity).await {
+                            Ok(response) => {
+                                let current_profile = response.profile;
+                                let update_profile = ProfileUpdate {
```
marcua, line 258: Given it's a `PATCH`, should we make it so the endpoint accepts a subset of the arguments and handles filling in the previous values on the server side? So I can send only a description or organization or links, and that's OK.

src/bin/ayb.rs
marcua: It seems like we sort of do that below with `get_many`, so this might instead be feedback on the documentation string, which made me think we can only accept one link, and it will overwrite all other links.

README.md
```diff
@@ -105,6 +105,15 @@ $ ayb client query marcua/test.sqlite "SELECT * FROM favorite_databases;"
  DuckDB     | 9 
 
 Rows: 3
+
```
marcua, line 108: Great doc updates! :)

src/http/structs.rs
```diff
@@ -128,36 +143,77 @@ pub struct EntityQueryResponse {
...
+            Cell::new("Links"),
+        ]));
+
+        table.add_row(Row::new(vec![
+            Cell::new(
+                &self
+                    .display_name
+                    .as_deref()
+                    .unwrap_or("null")
+                    .replace(',', "\\,"),
```
marcua, line 163: Throughout this function, why do we have to escape a backslash before the comma?

src/http/structs.rs
```diff
@@ -128,36 +143,77 @@ pub struct EntityQueryResponse {
...
+            ),
+            Cell::new(
+                &self
+                    .links
+                    .clone()
+                    .iter()
+                    .map(|v| v.url.as_str())
```
marcua, line 191: Also mention if it's verified, or nah? I don't feel strongly

src/hosted_db.rs
```diff
@@ -41,18 +42,12 @@ impl QueryResult {
         table
     }
 
-    pub fn generate_table(&self) -> Result<(), std::io::Error> {
```
marcua, line 44: do we need `generate_table` anymore?

src/lib.rs
```diff
@@ -2,9 +2,26 @@
...
+use prettytable::{format, Table};
+
 pub mod ayb_db;
 pub mod email;
 pub mod error;
 pub mod hosted_db;
 pub mod http;
 pub mod templating;
+
+pub trait FormatResponse {
```
marcua, line 14: Should this live in its own module like `formatting`?

And should we name it something like `RelationalFormatter` or `TabularFormatter` since it formats relations/tables?

tests/e2e.rs
```diff
@@ -402,5 +458,33 @@ fn client_server_integration(
         &format!("No queryable databases owned by {}", first_entity_0),
     )?;
 
+    update_profile(
```
marcua, line 461: Assuming you agree with my feedback on only having to update a subset of fields, should we test
* Update field 1, field 2
* Profile has field 1 / field 2
* Update field 2, field 3
* Profile has field 1 / new field 2 / field 3?

tests/e2e.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::too_many_arguments)]
```
marcua, line 1: Sanity-checking this

marcua (changes requested):
Amaaazing! This is really coming together! :)

Left some small thoughts on the CLI, and otherwise tiny review comments. Yay!!!

src/bin/ayb.rs
sofiaritz: You're right, that was left from the `POST` implementation :p
I'll update it :)

README.md
sofiaritz: Thank youu :)

src/http/structs.rs
sofiaritz: If someone's display name contains a comma and it's not escaped in the CSV format, CSV parsers will interpret the part after the comma as the next column.

src/hosted_db.rs
sofiaritz: We don't, I'll remove it

src/lib.rs
sofiaritz: I don't feel strongly, I'll move it to its own module. I like `TabularFormatter` :)

tests/e2e.rs
sofiaritz: The `update_profile` has more than 7 params, the ideal thing would be to create a struct, but because this is only for the E2E tests I don't think it matters much :p

sofiaritz:
Done! :)

Ready for another round :p

src/ayb_db/db_interfaces.rs
```diff
@@ -262,34 +262,63 @@ WHERE id = $1
...
-                .or_else(|err| match err {
-                    sqlx::Error::RowNotFound => Err(AybError::RecordNotFound {
-                        id: entity_id.to_string(),
-                        record_type: "entity".into(),
-                    }),
-                    _ => Err(AybError::from(err)),
-                })?;
+            async fn update_entity_by_id(&self, entity: &PartialEntity, entity_id: i32) -> Result<InstantiatedEntity, AybError> {
+                let mut query = QueryBuilder::new("UPDATE entity SET");
+                if let Some(description) = &entity.description {
```
marcua, line 267: I bet we can shorten the repetition on the next few branches by creating a vec/array of pairs like `("description", entity.description)` for any non-`None` item, and then looping over it.

src/ayb_db/models.rs
```diff
@@ -173,6 +173,37 @@ pub struct Entity {
     pub links: Option<Vec<Link>>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PartialEntity {
+    pub slug: Option<String>,
```
marcua, line 178: are things like slug/entity_type options, or only the display_name onward?

src/ayb_db/models.rs
```diff
@@ -173,6 +173,37 @@ pub struct Entity {
...
+    pub entity_type: Option<i16>,
+    pub display_name: Option<Option<String>>,
+    pub description: Option<Option<String>>,
+    pub organization: Option<Option<String>>,
+    pub location: Option<Option<String>>,
+    pub links: Option<Option<Vec<Link>>>,
+}
+
+impl Default for PartialEntity {
```
marcua, line 187: I didn't know about this --- I'm going to try to use the same approach on my WIP for #251! :)

src/bin/ayb.rs
```diff
@@ -1,9 +1,9 @@
 use ayb::ayb_db::models::{DBType, EntityType};
+use ayb::formatting::TabularFormatter;
```
marcua, line 2: I feel silly, bu where is this import used in the module?

src/bin/ayb.rs
```diff
@@ -149,7 +149,7 @@ async fn main() -> std::io::Result<()> {
                         .arg(arg!(--description <value> "New description").required(false))
                         .arg(arg!(--organization <value> "New organization").required(false))
                         .arg(arg!(--location <value> "New location").required(false))
-                        .arg(arg!(--link <value> "New URL, will overwrite previous ones").required(false))
+                        .arg(arg!(--link <value> "Adds a link, will overwrite previous ones").required(false))
```
marcua, line 152:
```suggestion
                        .arg(arg!(--link <value> "One of several links, all of which together will overwrite previous ones").required(false))
```

src/http/endpoints/update_profile.rs
```diff
@@ -25,50 +25,63 @@ pub async fn update_profile(
...
-        entity_type: instantiated_entity.entity_type,
-        display_name: profile.display_name,
-        description: profile.description,
-        organization: profile.organization,
-        location: profile.location,
-        links,
-    };
+    let mut partial = PartialEntity::new();
+    if let Some(display_name) = profile.display_name {
+        partial.display_name = Some(Some(display_name));
```
marcua, line 64: Two related questions:
* Why do we need `Some(Some(...))` here and below?
* Given the data coming off the wire contains `Option`s, can't we copy them from the struct being send from the client?

tests/e2e.rs
```diff
@@ -464,9 +466,37 @@ fn client_server_integration(
...
+
+    update_profile(
+        server_url,
+        &first_api_key0,
+        first_entity_0,
+        None,
+        Some("Entity 0 NEW description"),
         Some("Entity 0 organization"),
-        Some("Entity 0 location"),
-        vec!["http://example.com"],
```
marcua: Only because links are special/different from the other data types, should we still test them?

marcua (changes requested):
We're so close! Thank you as always for the great work! :)

src/ayb_db/models.rs
sofiaritz: Only the `display_name`, I forgot to delete that from something I tested :p

src/bin/ayb.rs
sofiaritz: When calling `generate_table` and `generate_csv` :)

You need the trait to be in scope in order to call its functions. This is also the case for std traits, but most of them are added to the [prelude](https://doc.rust-lang.org/core/prelude/rust_2021/index.html) and are imported automagically.

src/http/endpoints/update_profile.rs
sofiaritz: The idea of `Option<Option<T>>` is the following:
* Value is `None` &rarr; do not change anything.
* Value is `Some(None)` &rarr; set the value to `NULL`.
* Value is `Some(v)` &rarr; set the value to `v`.

We can't directly use the data coming off the wire because JSON doesn't have a way to represent an `Option<Option<T>>` :)
I could simplify this implementation using `map`, I think I'm going to update it :)

tests/e2e.rs
sofiaritz: You're right, I'll also test links

src/ayb_db/db_interfaces.rs
sofiaritz: That's a good idea :)

sofiaritz:
Done! I need to fix my local test environment to stop doing so many pushes :p
Suddenly I'm getting an "os error 2" error in the SQLite tests :woman_shrugging: 

Also, I ended up using `num_args(0..)` as you suggested earlier :)

src/bin/ayb.rs
marcua: This explains so many questions! :)

(It also opens up so many more, but about Rust and not this PR :P)

src/ayb_db/db_interfaces.rs
```diff
@@ -264,31 +264,34 @@ WHERE id = $1
...
+                let mut prev_to_links = false;
+                let pairs = vec![
+                    ("description", &entity.description),
+                    ("organization", &entity.organization),
+                    ("location", &entity.location),
+                    ("display_name", &entity.display_name),
+                ];
+
+                for (i, (key, current)) in pairs.iter().enumerate() {
+                    let Some(current) = current else {
```
marcua, line 276: ...and I learned something new :)

src/http/endpoints/update_profile.rs
marcua: Great explanation, and smart encoding! I have a slight preference for excluding the key entirely to represent no update, and then you only need a single level of `Option`. That said, I'll leave a comment below mentioning to document the encoding if you keep it this way.

src/ayb_db/models.rs
```diff
@@ -175,20 +175,22 @@ pub struct Entity {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct PartialEntity {
-    pub slug: Option<String>,
-    pub entity_type: Option<i16>,
     pub display_name: Option<Option<String>>,
```
marcua, line 178: While I prefer the single level of Option without needing to encode "no update," (using a map as you explained), if we keep it this way (it's ok if you do), copy some of the text from [this comment](https://github.com/marcua/ayb/pull/248#discussion_r1439326967) as a code comment so readers can understand how it works.

marcua (changes requested):
One tiny thing left, which is either a comment or a slightly different encoding. Thank you for this!! :)

src/ayb_db/db_interfaces.rs
sofiaritz: Haha, that's great! :)

src/http/endpoints/update_profile.rs
sofiaritz: I'm going to keep this struct as-is, because I don't really like the idea of throwing a `HashMap` around. But I've updated the implementation of the update entity endpoint and now HTTP clients can do exactly that:
* Do not include the key &rarr; `None`
* Key is `null` &rarr; `Some(None)`
* Key is `v` &rarr; Some(Some(v))

I'll add the docstring :)

src/bin/ayb.rs
sofiaritz: That's great! :)
Feel free to ask me privately any questions, I'll do my best to explain :p

sofiaritz:
Done! Ready for (hopefully) the last round! :)
:wrestling: :bell: :1st_place_medal: 

[Tests pass](https://github.com/sofiaritz/ayb/actions/runs/7396658582/job/20122239143)

marcua (approved):

marcua (approved):
Way to go! Another one for the books! :)

Merging now, and will release the next version soon. Let me know if you're blocked on the release.

Woohoo! Thank you so much!

sofiaritz:
Yay! :)
Don't worry about the release, there's no hurry :p

PR #259, opened 2024-01-06
Save API tokens and default server URL locally

### LINKED ISSUE #251: Make CLI save authentication tokens on registration/login
- [x] move all client-related code into `client`
- [x] introduce client/config.rs::ClientConfig struct
- [x] add read/write methods to serialize into json
- [x] update CLI to read/write file
- [x] add default_url handling

marcua:
Resolves #251


src/bin/ayb.rs
```diff
@@ -169,10 +176,23 @@ async fn main() -> std::io::Result<()> {
             Err(err) => println!("Error: {}", err),
         }
     } else if let Some(matches) = matches.subcommand_matches("client") {
+        let config_path = if let Some(path) = matches.get_one::<PathBuf>("config") {
+            path.clone()
+        } else {
+            home_dir()
+                .expect("can't determine home directory")
+                .join(".ayb.json")
```
sofiaritz, line 184: The standard (at least on Linux) is to create config files inside `$HOME/.config/$APP_NAME`.
In this case that would be `$HOME/.config/ayb/ayb.json`

src/client/config.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct ClientConfig {
+    version: u64,
+    pub authentication: HashMap<String, String>,
+    //default_url: Option<Url>
+}
+
+impl ClientConfig {
+    fn default_config() -> ClientConfig {
```
sofiaritz, line 15: As you may already know, perfect use case for [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) :p

src/bin/ayb.rs
marcua: Thank you! I switched over to `directories`. Hurray for standardizing across operating systems! :)

src/client/config.rs
marcua: Thank you and done!

src/bin/ayb.rs
```diff
@@ -169,174 +182,204 @@ async fn main() -> std::io::Result<()> {
...
+
+        let token = matches
+            .get_one::<String>("token")
+            .or(config.authentication.get(&url))
+            .cloned();
+        let client = AybClient {
+            base_url: url.to_string(),
+            api_token: token,
+        };
+        if let Some(matches) = matches.subcommand_matches("create_database") {
```
marcua, line 224: The changes below are just unindenting by one level since we removed a layer of `if` to get the URL above.

src/client.rs
```diff
@@ -0,0 +1,2 @@
+pub mod config;
```
marcua, line 1: Starting the move toward pulling things out of `http` and into `server` and `client`

src/client/config.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct ClientConfig {
+    version: u64,
+    pub authentication: HashMap<String, String>,
+    pub default_url: Option<String>,
+}
+
+impl ClientConfig {
+    pub fn new() -> ClientConfig {
```
sofiaritz, line 15: Should `new` receive the version as a param or should we leave it as is and change it in the future if needed?

src/client/config.rs
```diff
@@ -0,0 +1,51 @@
...
+        ClientConfig {
+            version: 1,
+            authentication: HashMap::new(),
+            default_url: None,
+        }
+    }
+
+    pub fn from_file(file_path: &PathBuf) -> Result<ClientConfig, std::io::Error> {
+        if file_path.exists() {
+            let file = File::open(file_path)?;
```
sofiaritz, line 25: You could use `fs::read_to_string()` here to simplify this function a bit

src/client/config.rs
```diff
@@ -0,0 +1,51 @@
...
+            let mut reader = BufReader::new(file);
+            return Ok(serde_json::from_reader(&mut reader)?);
+        }
+
+        Ok(ClientConfig::new())
```
sofiaritz, line 30: This should call `default` instead of `new`

src/client/config.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    pub fn to_file(&self, file_path: &PathBuf) -> Result<(), std::io::Error> {
+        create_dir_all(
+            file_path
+                .parent()
+                .expect("unable to determine parent of ayb configuration directory"),
+        )?;
+        let file = File::create(file_path)?;
```
sofiaritz, line 39: You could use `fs::write` to simplify this function

sofiaritz (changes requested):
I really like how you implemented this, great work! :)
Just some suggestions

README.md
```diff
@@ -116,6 +112,13 @@ $ ayb client profile marcua
  Adam Marcus  |             |              |          | http://marcua.net 
 ```
 
+Note that the command line also saved a configuration file (on Linux,
+to a place like `$HOME/.config/ayb/ayb.json`) for your convenience so
```
sofiaritz, line 116: Maybe adding three bullet points for Linux, MacOS and Windows would be useful for some people

sofiaritz (commented):
I forgot one comment :p
Also, would a command like `ayb client clean` to delete the config file be useful?

src/client/config.rs
marcua: Good question! My intuition is no, since the version is a function of the schema/format of the encoded JSON, and should stay close to it. I'm adding it here now in case we ever need to change the schema/format, so that we can detect an older version of the configuration file and migrate it. The caller shouldn't have any knowledge of this internal detail of the configuration file. For example, you wouldn't want a caller to say that the configuration format is a v2 format when it is in fact v1. Let me know if that makes sense.

src/client/config.rs
marcua: Done!

src/client/config.rs
marcua: Done! I in initially had the buffered reader/writers for performance reasons, but the configuration file is so tiny that I suspect it will hardly matter :).

src/client/config.rs
marcua: Done! Thank you!

README.md
marcua: Done!

marcua:
Thank you so much for the review @sofiaritz ! That simplified and clarified things! Let me know if my response on passing a version makes sense. Re: `ayb clean`, I haven't seen it in too many CLIs, but we can look out for use cases for it and add it if we encounter them. What do you think?

src/client/config.rs
sofiaritz: I think you're right, I'll mark this as resolved :)

sofiaritz (approved):
Great implementation! I think we can leave the `ayb client clean` command for another day :)

The idea of the `ayb client clean` command is to leave the file system like it was before `ayb` was installed and used. That way someone who wants to remove `ayb` from their system can do so without leaving anything behind.

marcua:
Thank you so much @sofiaritz! :)

I like the `ayb client clean` idea. A nice uninstaller for when you no longer want it! We can keep our eyes out for people who ask for it. Thank you again!

PR #269, opened 2024-01-11
Test for different casing on entity slugs, catch an error in the database interface, lowercase slugs before query

### LINKED ISSUE #266: Slugs with uppercase characters makes some endpoints fail
The entity retrieval endpoint returns `RecordNotFound` when the given slug contains any uppercase character.
I think we should disallow slugs with uppercase characters altogether to prevent confusion and possible abuse (`admin` ≠ `Admin` internally but may mislead people).

Users can change their display name anyways.
Thanks @Suguivy for discovering this issue!

marcua:
Resolves #266

### DIFF
--- a/src/ayb_db/db_interfaces.rs
+++ b/src/ayb_db/db_interfaces.rs
@@ -191,7 +191,14 @@ WHERE
                 .bind(entity_slug)
                 .bind(database_slug)
                 .fetch_one(&self.pool)
-                .await?;
+                .await
+                .or_else(|err| match err {
+                    sqlx::Error::RowNotFound => Err(AybError::RecordNotFound {
+                        id: format!("{}/{}", entity_slug, database_slug),
+                        record_type: "database".into(),
+                    }),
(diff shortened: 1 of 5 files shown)


src/ayb_db/db_interfaces.rs
```diff
@@ -191,7 +191,14 @@ WHERE
                 .bind(entity_slug)
                 .bind(database_slug)
                 .fetch_one(&self.pool)
-                .await?;
+                .await
+                .or_else(|err| match err {
+                    sqlx::Error::RowNotFound => Err(AybError::RecordNotFound {
```
marcua, line 196: I discovered a tiny bug when I added a test for retrieving databases with a case-sensitive entity slug. Adding it to this PR since the test discovered the issue :).

sofiaritz (approved):
I'll reply to the comment you left in the issue tomorrow when I have some free time :)
This seems like the right solution for now

marcua:
Hooray! Thank you @sofiaritz! :shipit:

PR #279, opened 2024-01-18
Return entity slug with confirmation endpoint

### LINKED ISSUE #260: Return entity slug with confirmation endpoint
This would allow frontends to redirect the user to their profile after logging in

marcua:
Resolves #260


sofiaritz (approved):
Yay!! Looks great :)

marcua:
Hooray! Thank you! :)

PR #284, opened 2024-01-21
Refactor: Move everything but structs from http to server module

marcua:


src/http/structs.rs
```diff
@@ -5,58 +5,6 @@ use crate::ayb_db::models::{
 use crate::formatting::TabularFormatter;
 use prettytable::{Cell, Row, Table};
 use serde::{Deserialize, Serialize};
-use url::Url;
```
marcua, line 8: This is the only "substantive" part of the refactor (moving AybConfig* into `server/config.rs`) --- everything else is directory renaming

Issue #275, opened 2024-01-16
REPL/interactive mode for multiple queries


Suguivy:
Right now, to do multiple queries in `ayb`, one must write `ayb query <database> <query>` for each query they want to execute in the database. Adding a REPL/interactive mode for multiple queries could help to execute multiple queries easier.

For example, instead of doing this:

```sh
sugui@yukari:~$ ayb client query sugui/test-from-cli "create table numbers(spelling varchar, num integer);"

Rows: 0
sugui@yukari:~$ ayb client query sugui/test-from-cli "insert into numbers values ('seven', 7)"

Rows: 0
sugui@yukari:~$ ayb client query sugui/test-from-cli "select * from numbers"
 spelling | num 
----------+-----
 seven    | 7 

Rows: 1
```

It could exist something like a `client interactive` option to enter an interactive mode where you can simply start to write queries into the same database.
```sh
sugui@yukari:~$ ayb client interactive sugui/test-from-cli 
sugui@sugui/test-from-cli> create table numbers(spelling varchar, num integer);

Rows: 0
sugui@sugui/test-from-cli> insert into numbers values ('seven', 7);

Rows: 0
sugui@sugui/test-from-cli> select * from numbers;
 spelling | num 
----------+-----
 seven    | 7 

Rows: 1
sugui@sugui/test-from-cli> ^C
sugui@yukari:~$
```

This is an improvement I told @sofiaritz about. She told me I could open an issue for that, so here it is :)

marcua:
Thank you for this excellent idea, @Suguivy! :) I would love to get this feature into `ayb`. Your proposal looks great to me. Two thoughts on this implementation
* It looks like we can rely on [rustyline](https://crates.io/crates/rustyline) do do the heavy lifting of reading a line of input with support for history, hotkeys, etc.
* Can you or @sofiaritz think of any reason why we need to give it its own CLI command? For example, do we want `ayb client interactive sugui/test-from-cli ` with an explicit `interactive` command, or should it be `ayb client query sugui/test-from-cli`, and if there's no query at the end of this command, it implicitly goes into interactive mode?

I'm quite excited about this idea, and would love to either help you contribute it if you wanted to write the code, or to write it and get your feedback on an early implementation. (And no problem if you wanted to share the idea but not more than that.) Do you have a preference?

Suguivy:
Thank you for the response @marcua! I will reply now to the two points you highlighted:

[...]

PR #286, opened 2024-01-26
Interactive mode for multiple queries

Suguivy:
Implements the interactive mode specified in #275

Feel free to do any corrections or suggestions before merging :)

### DIFF
--- a/src/bin/ayb.rs
+++ b/src/bin/ayb.rs
@@ -2,7 +2,6 @@ use ayb::ayb_db::models::{DBType, EntityType};
 use ayb::client::config::ClientConfig;
 use ayb::client::http::AybClient;
 use ayb::formatting::TabularFormatter;
-use ayb::hosted_db::QueryResult;
 use ayb::http::structs::{EntityDatabasePath, ProfileLinkUpdate};
 use ayb::server::config::{config_to_toml, default_server_config};
 use ayb::server::server_runner::run_server;
@@ -28,14 +27,23 @@ fn entity_database_parser(value: &str) -> Result<EntityDatabasePath, String> {
     }
 }
 
-fn display_query_result(query_result: &QueryResult, format: &OutputFormat) -> Result<(), std::io::Error> {
-    if !query_result.rows.is_empty() {
-        match format {
-            OutputFormat::Table => query_result.generate_table()?,
-            OutputFormat::Csv => query_result.generate_csv()?,
+async fn query_and_display(client: &AybClient, entity: &str, database: &str, query: &str, format: &OutputFormat) -> Result<(), std::io::Error> {
+    match client
+    .query(&entity, &database, query)
+    .await
+    {
+        Ok(query_result) => {
+            if !query_result.rows.is_empty() {
+                match format {
+                    OutputFormat::Table => query_result.generate_table()?,
+                    OutputFormat::Csv => query_result.generate_csv()?,
+                }
+            }
+            println!("\nRows: {}", query_result.rows.len());        }
+        Err(err) => {
+            println!("Error: {}", err);
         }
     }
-    println!("\nRows: {}", query_result.rows.len());
     Ok(())
 }
 
@@ -379,17 +387,7 @@ async fn main() -> std::io::Result<()> {
                 matches.get_one::<OutputFormat>("format"),
             ) {
                 if let Some(query) = matches.get_one::<String>("query") {
-                    match client
-                    .query(&entity_database.entity, &entity_database.database, query)
-                    .await
-                    {
-                        Ok(query_result) => {
-                            display_query_result(&query_result, format)?;
-                        }
-                        Err(err) => {
-                            println!("Error: {}", err);
-                        }
-                    }   
+                    query_and_display(&client, &entity_database.entity, &entity_database.database, query, format).await?;
                 } else {
                     println!("Launching an interactive session for {}/{}", entity_database.entity, entity_database.database);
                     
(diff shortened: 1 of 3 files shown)


src/bin/ayb.rs
```diff
@@ -360,26 +374,58 @@ async fn main() -> std::io::Result<()> {
...
+                        Err(err) => {
+                            println!("Error: {}", err);
+                        }
+                    }   
+                } else {
+                    println!("Launching an interactive session for {}/{}", entity_database.entity, entity_database.database);
+                    
+                    match DefaultEditor::new() {
+                        Ok(mut rl) => loop {
+                            let line = rl.readline(">>> ");
```
marcua, line 398: I like that the `psql` prompt tells you the name of the database you're connected to, e.g.,

```
$ psql -U postgres_user -d test_db -W -h localhost
Password: 
...
...
test_db=>
```

Would it be worth replacing `>>>` with something like `entity/database>` or `entity/database=>`?

src/bin/ayb.rs
```diff
@@ -360,26 +374,58 @@ async fn main() -> std::io::Result<()> {
...
-                                OutputFormat::Table => query_result.generate_table()?,
-                                OutputFormat::Csv => query_result.generate_csv()?,
+                    {
+                        Ok(query_result) => {
+                            display_query_result(&query_result, format)?;
+                        }
+                        Err(err) => {
+                            println!("Error: {}", err);
+                        }
+                    }
```
marcua, line 392: I like that you created a function `display_query_result` since its logic is used in two places. There's an opportunity to expand that idea: It looks like this entire block (line 382-392) is the same as the one on 407-417. Instead of `display_query_result`, you could have a function `query_and_display(client, entity, database, query, format)` that covers both blocks.

src/bin/ayb.rs
```diff
@@ -94,8 +108,8 @@ async fn main() -> std::io::Result<()> {
                              .value_parser(ValueParser::new(entity_database_parser))
                              .required(true)
                         )
-                        .arg(arg!(<query> "The query to execute")
-                             .required(true))
+                        .arg(arg!(<query> "The query to execute. If not provided, an interactive session to write queries will be launched")
```
marcua, line 111:
```suggestion
                        .arg(arg!(<query> "The query to execute. If not provided, an interactive session to write queries will be launched.")
```

src/bin/ayb.rs
```diff
@@ -360,26 +374,58 @@ async fn main() -> std::io::Result<()> {
...
+                            display_query_result(&query_result, format)?;
+                        }
+                        Err(err) => {
+                            println!("Error: {}", err);
+                        }
+                    }   
+                } else {
+                    println!("Launching an interactive session for {}/{}", entity_database.entity, entity_database.database);
+                    
+                    match DefaultEditor::new() {
```
marcua, line 396: I'm OK with you using `?` to avoid the extra level of nesting, but I don't feel strongly :).

src/bin/ayb.rs
```diff
@@ -360,26 +374,58 @@ async fn main() -> std::io::Result<()> {
...
+                        Ok(mut rl) => loop {
+                            let line = rl.readline(">>> ");
+                            match line {
+                                Ok(line) if line.is_empty() => {}
+                                Ok(query) => {
+                                    let result = rl.add_history_entry(query.as_str());
```
marcua, line 402: I'm similarly OK using ? or not checking the result here, since we're not saving the history to a file.

marcua:
Way to go @Suguivy! This is such a great contribution! I checked out the code to play with it, and it worked flawlessly.

I left some comments for you and am happy to discuss any of it. If it helps for me to make any of the edits directly let me know, but you've done a really nice job with the v1!

Thank you so much! :)

src/bin/ayb.rs
Suguivy: I have seen that other arguments/flags/subcommands descriptions don't end with a dot. Should we proceed with that change anyways?

src/bin/ayb.rs
Suguivy: Yeah I think it's a good idea, it's a more informative prompt that way :)

I committed that change.

src/bin/ayb.rs
Suguivy: Yeah it would be better to avoid repeating code. Committed the change!

src/bin/ayb.rs
Suguivy: We could add an `.unwrap()` there to avoid checking the result, it's not necessary to check imo.

I don't know how can we add a `?` there because seems that there is a mismatch on the different error types :(

src/bin/ayb.rs
Suguivy: Hmm, I don't know where we can exactly use a `?` there.

Suguivy:
Thank you for the feedback @marcua :)

I committed some changes for some of the improvements you pointed out. Feel free to do remaining edits directly if you want!

src/bin/ayb.rs
marcua: Thank you for pushing for consistency! :) While `ayb` doesn't have copy guidelines, the rule I've used in other projects/presentations is that if you have multiple sentences, to end the last one in a period. Since we've got `The query to execute.`, the second sentence would end in a period. Not a strict requirement, but figured it would be good here. I'll make the change after merging :).

src/bin/ayb.rs
marcua: I suspect we can drop the `match` and call `DefaultEditor::new()?`, but will try it after merging :)

src/bin/ayb.rs
marcua: Ugh error types. :( I'll see if there's a trick I can play with it, or if it's not worth it and perhaps an unwrap in the client is fine.

marcua:
Wonderful work @Suguivy ! :)

I've responded to your great questions, and will take your advice to merge and edit after merging. I'll share the follow-up PR with you.

Congratulations on your first contribution (hopefully of many) to `ayb`! Thank you for the great idea and the implementation! :)

Suguivy:
Thank you both @marcua and @sofiaritz to help me and give me the opportunity to collaborate :)

Issue #14, opened 2022-11-26
Back up databases to S3-compatible storage


marcua:
There are many options for backing up SQLite to S3. The one that most directly focuses on continuous streaming backups and restore (as opposed to maintaining read replicas or multiple writers, etc.) is [litestream](https://litestream.io/getting-started/), but there are others like [libSQL's bottomless replication work-in-progress](https://github.com/tursodatabase/libsql/tree/main/bottomless).

We discuss alternatives and why we aren't (yet) considering them below, but for now, we will go with [simple scheduled backups](https://litestream.io/alternatives/cron/).
* In `ayb.toml`, you will configure: S3-compatible storage details, frequency of backup, and how many backups to keep before pruning old ones.
* `ayb` will have a Snapshot data model that can store details on database snapshots (when we took it, what the hash of the database directory was, whether it was a scheduled or manual snapshot, the path to the backup on S3-compatible storage, etc.)
* On the `ayb` server, we'll run a periodic task to check if a database has changed recently (based on the hash of its database directory, but perhaps there's a way to query for the most recent transaction ID instead) and if it has, create a new snapshot.
* The backup in SQLite will happen via [VACUUM](https://www.sqlite.org/lang_vacuum.html), which will reduce file size and won't potentially run forever under high write loads like the [online backup API](https://www.sqlite.org/backup.html).
* The advantages of this approach are:
  * Simplest implementation path.
  * No dependence on other tools that are being rapidly developed and don't have documented APIs (yet).
  * Due to the simplicity and generality of the approach, it shouldn't be hard to apply it to other database types (e.g., DuckDB) that have less developed stream-the-WAL-to-S3 functionality.
  * I suspect that in practice, most databases will be write-dormant most of the time given the target use cases, so this might require less overhead than tools that constantly monitor a database log for changes.
* The disadvantages of the approach are:
  * Potentially more work per backup: if you only update a row, we still VACUUM and export the entire database.
  * Periodic snapshots have a larger window for data loss on the order of tens of minutes to an hour, whereas shipping the write-ahead log typically results in only seconds of data loss.
  * Related to the previous reason, implementing things like multi-node `ayb` support with automatic failover will be difficult in the future in cases where a node becomes inaccessible and you're not sure how many minutes/tens of minutes of backups are unsynced.

The plan
[...]

PR #317, opened 2024-03-16
Introduce snapshot model, storage manager, and refactor model macros for reuse

marcua:
Related to #14

This PR introduces an API for listing and uploading gzipped snapshots with metadata to an S3-like storage system. No integration tests yet---that is coming in the next PR.

### DIFF
--- a/src/server/snapshots/storage.rs
+++ b/src/server/snapshots/storage.rs
@@ -1,68 +1,132 @@
 use crate::error::AybError;
 use crate::server::config::AybConfigSnapshots;
-use crate::server::snapshots::models::{InstantiatedSnapshot, ListSnapshotResult};
-use s3::bucket::Bucket;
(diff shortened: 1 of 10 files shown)


src/server/snapshots/storage.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+    fn db_path(self, entity_slug: &str, database_slug: &str, final_path: &str) -> String {
+        format!(
+            "{}/{}/{}/{}",
+            self.path_prefix, entity_slug, database_slug, final_path
+        )
+    }
+
+    // switch to rusty: https://docs.rs/rusty-s3/latest/rusty_s3/
+    // or https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html
```
marcua, line 44: start here next time

PR #372, opened 2024-05-20
Refactor end-to-end tests into smaller groupings

marcua:
* Ahead of adding tests for two new endpoints, I found the e2e tests that previously all lived in a single file to be unruly
* This PR breaks the tests into three groups of tests, and moves utility functions into their own module
* The e2e.rs entrypoint is now a lot smaller
* Meta-testing: after finishing this refactor, I broke an assertion at the bottom of each file to ensure the issue was still caught


tests/e2e_tests/registration_tests.rs
```diff
@@ -0,0 +1,229 @@
...
+        Ok(captures.get(2).map_or("", |m| m.as_str()).to_string())
+    } else {
+        Err(AybError::Other {
+            message: "No API key".to_string(),
+        })
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct EmailEntry {
```
marcua, line 27: Not pub

PR #405, opened 2024-07-02
Ability to restore from a snapshot

marcua:
This PR introduces the ability to restore a database from a snapshot stored in S3-compatible storage as part of #14.
* It introduces a `retrieve_snapshot` endpoint on the client and server, including end-to-end tests of that experience
* In order to do this in a standard way, it introduces some new path convenience functions
* It also renames some previously non-specific properties of snapshots (in particular, `snapshot_name` -> `snapshot_id`)
* Finally, it resolves several outstanding TODOs in the code that were in tangentially related code

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ prefixed-api-key = { version = "0.2.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
 url = { version = "2.5.1", features = ["serde"] }
 urlencoding = { version = "2.1.3" }
+uuid = { version = "1.9.1", features = ["v7"] }
 walkdir = { version = "2.5.0" }
 
 [dev-dependencies]
--- a/src/hosted_db/paths.rs
+++ b/src/hosted_db/paths.rs
@@ -1,7 +1,11 @@
 use crate::error::AybError;
 use std::fs;
+use std::os::unix::fs::symlink;
 use std::path::{Path, PathBuf};
+use uuid::{timestamp::context::ContextV7, Timestamp, Uuid};
 
+pub const CURRENT: &str = "current";
+pub const CURRENT_TMP: &str = "current.tmp";
 const DATABASES: &str = "databases";
 const SNAPSHOTS: &str = "snapshots";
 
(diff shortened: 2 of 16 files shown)


src/hosted_db/paths.rs
```diff
@@ -98,3 +142,29 @@ pub fn pathbuf_to_parent(path: &Path) -> Result<PathBuf, AybError> {
...
+/// current path to it) and, if a previous database existed as the
+/// current database, delete it.
+pub fn set_current_database_and_clean_up(new_path: &Path) -> Result<(), AybError> {
+    let mut current_db_path = pathbuf_to_parent(new_path)?;
+    let mut current_tmp_db_path = current_db_path.clone();
+    current_db_path.push(CURRENT);
+    current_tmp_db_path.push(CURRENT_TMP);
+    let previous_database_path = fs::canonicalize(current_db_path.clone());
+
+    // TODO(marcua): This `symlink` call is the Unix API. Consider
```
marcua, line 156: do

src/server/endpoints/restore_snapshot.rs
```diff
@@ -0,0 +1,63 @@
...
+        if let Some(ref snapshot_config) = ayb_config.snapshots {
+            // TODO(marcua): In the future, consider quiescing
+            // requests to this database during the process, and
+            // locking so that only one snapshot per database can be
+            // restored at a time.
+
+            // Retrieve the snapshot, move it to the active databases
+            // directory, and set it as the current active database.
+            let snapshot_storage = SnapshotStorage::new(snapshot_config).await?;
+            // TODO(marcua): Consider passing a path to
```
marcua, line 38: do

Cargo.toml
```diff
@@ -46,6 +46,7 @@ prefixed-api-key = { version = "0.2.0", features = ["sha2"]}
 prettytable-rs = { version = "0.10.0"}
```
marcua, line 46: Fix clippy

src/hosted_db/paths.rs
```diff
@@ -21,35 +25,75 @@ pub fn database_parent_path(data_path: &str, create_path: bool) -> Result<PathBu
...
+/// Returns a path for a new database directory for storing
+/// `{entity_slug}/{database_slug}`. The format for this path is
+/// `{data_path}/{entity_slug}/{database_slug}/{time_sortable_uuid}/`.
+pub fn new_database_path(
     entity_slug: &str,
     database_slug: &str,
     data_path: &str,
-    create_database: bool,
 ) -> Result<PathBuf, AybError> {
+    let uuid = Uuid::new_v7(Timestamp::now(ContextV7::new()));
```
sofiaritz, line 36: I'm interested, why do you use UUIDv7? :)

src/hosted_db/paths.rs
```diff
@@ -21,35 +25,75 @@ pub fn database_parent_path(data_path: &str, create_path: bool) -> Result<PathBu
...
-        .iter()
-        .collect();
-    if create_database {
-        if let Err(e) = fs::create_dir_all(&path) {
-            return Err(AybError::Other {
-                message: format!("Unable to create entity path for {}: {}", entity_slug, e),
-            });
-        }
-    }
+    let path: PathBuf = [
```
sofiaritz, line 40: This pattern is great!!

marcua: Aww thanks! I struggled for a while with it as a single file given the fact that there are journals, etc., and finally settled on every DB/iteration of a DB living in a directory :).

src/hosted_db/paths.rs
marcua: Great question! Sorry for the delay in replying---I created the PR and went on vacation :).

In practice, several versions of UUID would work in this specific situation, but here was my reasoning:
* The big downside of v7/time-sortable UUIDs is that they might leak sensitive details, namely the quantity/frequency of UUIDs that are created. If you're trying to hide, for example, how many customers you have, UUIDv7 will leak details similar to how serial integer primary keys do. We don't externally expose the UUIDs of these directories, and even if we did leak them, I couldn't think of a reason why the creation timestamp of a database's folder is sensitive information.
* One very weak benefit of v7 is, because it's time-sortable, if you ever end up in a corrupted state and need to manually dig through the directory structure, being able to determine which folder was created before/after other folders might help you recover the most recent snapshot/version of a database. (This information would also be available in filesystem and S3 metadata, so it's only a minor benefit.)
* Of all of the time-sortable UUID versions, v7 has the benefit that, when stored as bytes in PostgreSQL (or likely any B-Tree index), there is better temporal locality if you're inserting roughly in time order, which has performance and space benefits [1, 2]. We're not storing these particular UUID instances in a database, but if we're using any time-sortable UUID in the project and v7 outperforms the others when stored in a database, we may as well use that same version of UUID in other parts of the project all other things being equal.

[1] https://ardentperf.com/2024/02/03/uuid-benchmark-war/
[2] https://maciejwalkowiak.com/blog/postgres-uuid-primary-key/

src/server/snapshots.rs
```diff
@@ -88,12 +112,11 @@ pub async fn snapshot_database(
...
     match ayb_db.get_database(&entity_slug, &database_slug).await {
-        Ok(db) => {
-            println!("Hashing {} {}", entity_slug, database_slug);
+        Ok(_db) => {
             // TODO(marcua): Implement hashing. `.sha3sum --schema` is
             // only available at the SQLite command line since it's a
             // dot command.
-            let db_path = database_path(&entity_slug, &database_slug, &config.data_path, false)?;
+            let db_path = current_database_path(&entity_slug, &database_slug, &config.data_path)?;
             // TODO(marcua): Do better than "temporary"
```
marcua, line 120: do, with uuidv7

src/hosted_db/paths.rs
sofiaritz: Thank you for the thorough response! I don't know much about UUIDs aside from v4, so this is really insightful :)

PR #418, opened 2024-07-28
Don't upload snapshot if it's already in S3, prune snapshots once list gets too long

marcua:
This PR introduces optimizations to the hash-based snapshots introduced as part of #14:
* After taking a snapshot, it lists snapshots already uploaded to S3-compatible storage
* If the `blake3` hash of the snapshot it just took already exists in the snapshot list, it doesn't upload it again
* If it does upload a new snapshot, it checks whether the number of snapshots now exceeds `config.snapshots.automation.max_snapshots`, and it does, it prunes the oldest snapshots from S3-compatible storage.
* Future work: avoid taking a snapshot and hashing it at all if the modification times of the database files don't differ from previously cached ones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0.120", features = ["raw_value"] }
 serde_repr = { version = "0.1.19" }
 sqlx = { version = "0.6.3", features = ["chrono", "postgres", "runtime-actix-native-tls", "sqlite"] }
-time = { verson = "0.3.36" }  # Remove line once Rust > 1.80 is out: https://github.com/rust-lang/rust/pull/124736
 toml = { version = "0.8.14" }
 tokio = { version = "1.38.0", features = ["macros", "process", "rt"] }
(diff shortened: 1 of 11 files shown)


src/server/snapshots/storage.rs
```diff
@@ -114,6 +107,24 @@ impl SnapshotStorage {
         Ok(())
     }
 
+    /// Deletes all files in this bucket for
+    /// `entity_slug/database_slug`. Currently only used in tests to
+    /// get to a predictable starting point.
+    pub async fn delete_all_snapshots(
```
marcua, line 113: inline since it's only used in one place

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0.120", features = ["raw_value"] }
 serde_repr = { version = "0.1.19" }
 sqlx = { version = "0.6.3", features = ["chrono", "postgres", "runtime-actix-native-tls", "sqlite"] }
+time = { version = "0.3.36" }
```
marcua, line 43: due to https://github.com/time-rs/time/issues/681

PR #511, opened 2025-01-19
Investigate and fix CI create DB error

marcua:
Starting with Ubuntu 24.x, nsjail can no longer work with the default security permissions. We'll have to consider a solution for the long-term, but in the short-term, the solution on CI is to [turn off some AppArmor protections](https://github.com/google/nsjail/issues/236).

In debugging this issue, I realized that we weren't properly handling the case where the isolated query runner was returning an error that couldn't be parsed as an `AybError`, so resolved that as well.


marcua:
It's this:

image


Which is [this](https://github.com/google/nsjail/issues/236)

PR #553, opened 2025-03-10
[WIP] Frontend database details/query support

### LINKED ISSUE #544: Implement database / query interface

marcua:
Resolves #544


src/server/ui_endpoints/database.rs
```diff
@@ -0,0 +1,507 @@
+use crate::ayb_db::db_interfaces::AybDb;
```
marcua, line 1: Notes after seeing AI-generated output
- src/server/api_endpoints/entity_details has to return the query access level the user has to this database
- src/server/ui_endpoints/database.rs should have one endpoint. `get` to that endpoint returns a page with breadcrumbs navigation, metadata about the database, and three tabs: query, sharing (not implemented yet), and snapshots (not implemented yet). The query tab is the default tab, and contains an empty query textarea with a query button. The query/sharing/snapshot tabs will eventually live in different endpoints that all fill in the tab details in the same template.
- when you run the query, it `post`s to /{entity}/{database}/query with parameters for the query and the result type, which can be one of: html, csv, or json.
- in html mode, the query template is returned with a table below the query (or an error if one occurred during the query). It issues the query using the ayb_client's query function and returns HTML. The template only displays options to download CSV or JSON upon successful query.
- in CSV or JSON mode, the results are returned in those formats (or as errors if the query encountered an error).

marcua:
Compiles and works jankily. A few things I've noticed
* Tabs don't work
* Query isn't actually using HTMX, but is rendering on a new screen

marcua:
HTMX is functional! Running list
* Tabs don't work
* Clean up work specified in TODO in the HTML script include
* Better handle empty results and action buttons.
* Aesthetic pass over this screen.

marcua:
Tabs work! Build error fixed! Running list
* Clean up work specified in TODO in the HTML script include
* Better handle empty results and action buttons.
* Aesthetic pass over this screen.
* Sanity-check the other tabs

marcua:
Cleaned up HTML imports! Running list
* Consider `integrity` on all script/link tags
* Consider making HTMX import only on necessary (database) pages after looking at relative size of library
* Better handle empty results and action buttons.
* Aesthetic pass over this screen.
* Sanity-check the other tabs.

marcua:
We now handle errors, empty results, and consistent download buttons! I also did an aesthetic pass over the query tab.  Running list:
* Test/understand pagination.
* Sanity-check the other tabs.
* Clean up all of the tailwindisms that don't apply
* Code review

marcua:
Inserted rows for testing [1].  Running list:
* Fix pagination bug (all rows show up on first page even if more than 50, and previous/next buttons don't do anything.
* Clean up all of the tailwindisms that don't apply
* Code review

[1] How to insert ~150 rows:
```
INSERT INTO favorite_databases (name, score)
WITH RECURSIVE numbers(n) AS (
    SELECT 1
    UNION ALL
    SELECT n + 1 FROM numbers WHERE n < 150
)
SELECT 
    'Database ' || n,  -- Example value for column1
    7              -- Example value for column2
FROM numbers;
```

marcua:
Pagination (limiting result set, adding a note about pushing pagination into query in a separate issue #581) and query interface/result interface cleanup done! Time for code review/removing tailwindisms!

src/server/ui_endpoints/database.rs
```diff
@@ -0,0 +1,153 @@
...
+            r#"<li><a class="px-4 pb-3 pt-2" href="">Sharing</a></li>
+               <li><a class="px-4 pb-3 pt-2" href="">Snapshots</a></li>"#
+        } else {
+            ""
+        }
+    );
+
+    // Create query interface based on access level
+    let query_interface = match database_response.highest_query_access_level {
+        None => r#"<div class="uk-card p-4 bg-red-50 border border-red-200 rounded mb-4">
```
marcua, line 48: Stopped reviewing here --- can probably replace w/ uk-alert-destructive

PR #612, opened 2025-06-14
Introduce reserved username list and test for existing paths on registration

### LINKED ISSUE #530: Implement username registration & path routing precedence to allow /{username}/{database} paths
Summarized here: https://bsky.app/profile/marcua.net/post/3lhyq2qgaxk2n

marcua:
Resolves #530


src/server/api_endpoints/register.rs
```diff
@@ -20,6 +21,14 @@ async fn register(
     let entity = get_lowercased_header(&req, "entity")?;
     let email_address = get_lowercased_header(&req, "email-address")?;
     let entity_type = get_required_header(&req, "entity-type")?;
+
+    // Validate username for route conflicts
+    let base_url = ayb_config
```
marcua, line 26: We do this in web config --- can we share or use that function?

PR #615, opened 2025-06-25
Support for registration messaging backends other than SMTP (e.g., local file)

### LINKED ISSUE #217: Add a developer-friendly way to use registration without an SMTP server

marcua:
Resolves #217 and generally makes integration testing easier.

As part of this, resolved two tangential issues that made tests fail:
* LocalStack (but really the Colima/Docker stack on macOS) was causing by migrating to MinIO. The network setup was unstable on Colima/macOS. Seeing the fix (`--network host`), it might be unnecessary to have left LocalStack, but MinIO also seems like a more direct solution to the S3-compatible storage problem than the more general-purpose LocalStack.
* Various race conditions between (asynchronous) snapshots and test assertions made tests fail relatively frequently. Added sleeps between inserts, updates, and snapshot restores to allow predictability of snapshot count/state between modifications.


tests/e2e_tests/registration_tests.rs
```diff
@@ -115,18 +119,19 @@ pub fn test_registration(
...
-    let first_token1 = extract_token(&entries[1])?;
-    let entries = parse_smtp_log(&format!(
-        "tests/smtp_data_{}/e2e-another@example.org",
-        smtp_port
-    ))?;
-    assert_eq!(entries.len(), 2);
-    let first_token2 = extract_token(&entries[0])?;
-    let second_token0 = extract_token(&entries[1])?;
+    // Check that four emails were received total
+    assert_eq!(count_emails_in_file(config_path)?, 4);
```
marcua, line 123: Remove count_emails_in_file and all references to it

tests/email_helpers.rs
```diff
@@ -0,0 +1,89 @@
...
+                "\tayb client confirm abc123token".to_string(),
+            ],
+        }];
+
+        let token = extract_token_from_emails(&emails);
+        assert_eq!(token, Some("abc123token".to_string()));
+    }
+
+    #[test]
+    fn test_clear_email_file() {
```
marcua, line 81: Remove test, remove tempfile dependency

tests/e2e_tests/snapshot_tests.rs
```diff
@@ -161,6 +162,7 @@ pub async fn test_snapshots(
         "table",
         " the_count \n-----------\n 2 \n\nRows: 1",
     )?;
+    thread::sleep(time::Duration::from_secs(4));
 
     // There are 3 max_snapshots, so let's
```
marcua, line 167: Update comment

PR #671, opened 2025-10-06
Introduce environment variable overrides to configuration parameters

marcua:


src/server/config.rs
```diff
@@ -153,18 +152,40 @@ pub fn default_server_config() -> AybConfig {
 }
 
 pub fn read_config(config_path: &PathBuf) -> Result<AybConfig, AybError> {
-    let contents = fs::read_to_string(config_path).map_err(|err| AybError::ConfigurationError {
+    // Build layered configuration:
+    // 1. Start with TOML file (optional - won't error if missing)
+    // 2. Overlay environment variables with AYB_ prefix
```
marcua, line 157:
```suggestion
    // Overlay environment variables with AYB_ prefix on top of variables defined in TOML.
```

PR #752, opened 2026-04-19
Replace nsjail with landlock & setrlimit

marcua:


src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,173 @@
...
+///
+/// Configurable per-database limits and per-process CPU/thread
+/// limitation is future work.
+pub fn apply_sandbox(db_path: &Path) -> Result<(), AybError> {
+    #[cfg(target_os = "linux")]
+    {
+        apply_landlock_restrictions(db_path)?;
+        apply_resource_limits()?;
+    }
+    #[cfg(not(target_os = "linux"))]
```
marcua: Putting the warning here will end up printing in the daemon, no? We want to print once on server startup.

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,173 @@
...
+            ),
+        });
+    }
+    Ok(())
 }
 
-/// Build command for running daemon without isolation
-pub fn build_direct_command(db_path: &PathBuf) -> Result<tokio::process::Command, AybError> {
+/// Build command for running the query daemon. The daemon always
+/// sandboxes itself at startup via `apply_sandbox`.
```
marcua:
```suggestion
/// Build command for running the query daemon.
```

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,173 @@
...
+pub fn build_daemon_command(db_path: &PathBuf) -> Result<tokio::process::Command, AybError> {
```
marcua, line 236: This is called by main ayb binary, whereas the functions above are called by the query_daemon. For binary size, is it worth splitting into different modules, or does the compiler handle this for us?

src/hosted_db/sqlite.rs
```diff
@@ -94,16 +93,14 @@ pub fn query_sqlite(
     })
 }
 
-/// Run `query` against the database at `path`, either with or without isolation.
-pub async fn potentially_isolated_sqlite_query(
+/// Run `query` against the database at `path` via the sandboxed query
+/// daemon. The daemon applies Landlock + setrlimit isolation to itself
+/// at startup; see `src/hosted_db/sandbox.rs`.
```
marcua:
```suggestion
/// Run `query` against the database at `path` via the sandboxed query
/// daemon.
```

src/hosted_db/daemon_registry.rs
```diff
@@ -116,35 +115,24 @@ impl DaemonRegistry {
...
-        };
+    async fn spawn_daemon(&self, db_path: &PathBuf) -> Result<DaemonHandle, AybError> {
+        let mut cmd = build_daemon_command(db_path)?;
 
         // Spawn the process with piped stdin/stdout
         let mut child = cmd
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
-            .stderr(Stdio::null())
+            .stderr(Stdio::inherit())
```
marcua: Why the switch from null to inherit?

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,172 @@
...
+///
+/// Configurable per-database limits and per-process CPU/thread
+/// limitation is future work.
+pub fn apply_sandbox(db_path: &Path) -> Result<(), AybError> {
+    #[cfg(target_os = "linux")]
+    {
+        apply_landlock_restrictions(db_path)?;
+        apply_resource_limits()?;
+    }
+    #[cfg(not(target_os = "linux"))]
```
marcua: We already print this in the runner, no?

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,240 @@
...
+
+/// Isolation capabilities available on the current host, determined at
+/// server startup. The server prints a single status line based on this
+/// so operators know what protection they actually have.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IsolationStatus {
+    /// Linux 6.7+: Landlock filesystem + network + setrlimit.
+    Full,
+    /// Linux 5.13–6.6: Landlock filesystem + setrlimit (no network).
+    FilesystemOnly,
```
marcua:
```suggestion
    FilesystemAndRlimitOnly,
```

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,240 @@
...
+    }
+}
 
-    Ok(cmd)
+/// Parse the running kernel's `major.minor` version from
+/// `/proc/sys/kernel/osrelease`. Returns None if the file is missing
+/// or unparseable.
+#[cfg(target_os = "linux")]
+fn kernel_version() -> Option<(u32, u32)> {
+    let release = std::fs::read_to_string("/proc/sys/kernel/osrelease").ok()?;
```
marcua, line 46: Works on all modern Linuxes?

src/hosted_db/sandbox.rs
```diff
@@ -1,86 +1,240 @@
...
+///
+/// Configurable per-database limits and per-process CPU/thread
+/// limitation is future work.
+pub fn apply_sandbox(db_path: &Path) -> Result<(), AybError> {
+    #[cfg(target_os = "linux")]
+    {
+        apply_landlock_restrictions(db_path)?;
+        apply_resource_limits()?;
+    }
+    #[cfg(not(target_os = "linux"))]
```
marcua: do we need this check at all?

src/bin/ayb_query_daemon.rs
```diff
@@ -15,27 +16,37 @@ struct QueryRequest {
...
 /// And writes line-delimited JSON responses to stdout.
 ///
-/// This command can be run inside a sandbox that isolates
-/// parallel invocations from accessing each other's data, memory,
-/// and resources. That sandbox can be found in src/hosted_db/sandbox.rs.
+/// The daemon always applies Landlock filesystem and network restrictions
+/// plus resource limits (setrlimit) to itself before processing any
+/// queries. On platforms or kernels where Landlock is unavailable, a
+/// loud warning is printed to stderr and the daemon runs without
+/// isolation. See src/hosted_db/sandbox.rs.
```
marcua: This reads like the daemon prints the warning, or that the daemon always applies restrictions/limits. In practice it tries to apply as much of the sandbox as it can, and ayb warns you prominently about the elements of the sandbox it can't apply.

PR #764, opened 2026-04-26
Allow anonymous read access to entity profiles and public databases

marcua:
Anonymous viewers can now reach `/v1/entity/{slug}` and `/v1/{entity}/{db}/details` (and the corresponding `/{entity}` and `/{entity}/{db}` UI pages) without a bearer token, but only see databases whose `public_sharing_level` is `read-only` or `fork`. Query, fork, write, and management endpoints remain authenticated.

The `/v1` scope is wrapped with a new `optional_entity_validator` middleware that passes anonymous requests through (no entity inserted into request extensions) but still rejects invalid / revoked / expired bearer tokens — a bad token is never silently downgraded to anonymous. Auth-required endpoints continue to call `unwrap_authenticated_entity`, which now returns a new `AybError::Unauthorized` variant that produces a 401.

Permission helper signatures are unchanged. A new
`is_publicly_discoverable(&database)` predicate inspects only the public sharing level, and the two relaxed handlers branch on whether they have an authenticated entity to choose between `can_discover_database` and `is_publicly_discoverable`.

### DIFF
--- a/src/client/http.rs
+++ b/src/client/http.rs
@@ -34,6 +34,15 @@ impl AybClient {
         }
     }
 
+    fn add_optional_bearer_token(&self, headers: &mut HeaderMap) {
+        if let Some(api_token) = &self.api_token {
+            headers.insert(
+                HeaderName::from_static("authorization"),
+                HeaderValue::from_str(format!("Bearer {api_token}").as_str()).unwrap(),
+            );
+        }
+    }
+
     async fn handle_response<T: DeserializeOwned>(
         &self,
         response: reqwest::Response,
@@ -160,7 +169,7 @@ impl AybClient {
 
     pub async fn entity_details(&self, entity: &str) -> Result<EntityQueryResponse, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_optional_bearer_token(&mut headers);
 
         let response = reqwest::Client::new()
             .get(self.make_url(format!("entity/{entity}")))
@@ -274,7 +283,7 @@ impl AybClient {
         database: &str,
     ) -> Result<DatabaseDetails, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_optional_bearer_token(&mut headers);
 
         let response = reqwest::Client::new()
             .get(self.make_url(format!("{entity}/{database}/details")))
--- a/src/client/http.rs
+++ b/src/client/http.rs
@@ -19,27 +19,20 @@ impl AybClient {
         format!("{}/v1/{}", self.base_url, endpoint)
     }
 
-    fn add_bearer_token(&self, headers: &mut HeaderMap) -> Result<(), AybError> {
-        if let Some(api_token) = &self.api_token {
-            headers.insert(
-                HeaderName::from_static("authorization"),
-                HeaderValue::from_str(format!("Bearer {api_token}").as_str()).unwrap(),
-            );
-            Ok(())
-        } else {
-            Err(AybError::Other {
+    fn add_bearer_token(&self, headers: &mut HeaderMap, optional: bool) -> Result<(), AybError> {
+        match &self.api_token {
+            Some(api_token) => {
+                headers.insert(
+                    HeaderName::from_static("authorization"),
+                    HeaderValue::from_str(format!("Bearer {api_token}").as_str()).unwrap(),
+                );
+                Ok(())
+            }
+            None if optional => Ok(()),
+            None => Err(AybError::Other {
                 message: "Calling endpoint that requires client API token, but none provided"
                     .to_string(),
-            })
-        }
-    }
-
-    fn add_optional_bearer_token(&self, headers: &mut HeaderMap) {
-        if let Some(api_token) = &self.api_token {
-            headers.insert(
-                HeaderName::from_static("authorization"),
-                HeaderValue::from_str(format!("Bearer {api_token}").as_str()).unwrap(),
-            );
+            }),
         }
     }
 
@@ -120,7 +113,7 @@ impl AybClient {
             HeaderName::from_static("public-sharing-level"),
             HeaderValue::from_str(public_sharing_level.to_str()).unwrap(),
         );
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         let response = reqwest::Client::new()
             .post(self.make_url(format!("{entity}/{database}/create")))
@@ -138,7 +131,7 @@ impl AybClient {
         database: &str,
     ) -> Result<SnapshotList, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         let response = reqwest::Client::new()
             .get(self.make_url(format!("{entity}/{database}/list_snapshots")))
@@ -169,7 +162,7 @@ impl AybClient {
 
     pub async fn entity_details(&self, entity: &str) -> Result<EntityQueryResponse, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_optional_bearer_token(&mut headers);
+        self.add_bearer_token(&mut headers, true)?;
 
         let response = reqwest::Client::new()
             .get(self.make_url(format!("entity/{entity}")))
@@ -188,7 +181,7 @@ impl AybClient {
         query: &str,
     ) -> Result<QueryResult, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         let response = reqwest::Client::new()
             .post(self.make_url(format!("{entity}/{database}/query")))
@@ -238,7 +231,7 @@ impl AybClient {
         snapshot_id: &str,
     ) -> Result<(), AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         let response = reqwest::Client::new()
             .post(self.make_url(format!("{entity}/{database}/restore_snapshot")))
@@ -257,7 +250,7 @@ impl AybClient {
         profile_update: &HashMap<String, Option<String>>,
     ) -> Result<(), AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         headers.insert(
             "Content-Type",
@@ -283,7 +276,7 @@ impl AybClient {
         database: &str,
     ) -> Result<DatabaseDetails, AybError> {
         let mut headers = HeaderMap::new();
-        self.add_optional_bearer_token(&mut headers);
+        self.add_bearer_token(&mut headers, true)?;
 
         let response = reqwest::Client::new()
             .get(self.make_url(format!("{entity}/{database}/details")))
@@ -302,7 +295,7 @@ impl AybClient {
         public_sharing_level: &PublicSharingLevel,
     ) -> Result<(), AybError> {
         let mut headers = HeaderMap::new();
-        self.add_bearer_token(&mut headers)?;
+        self.add_bearer_token(&mut headers, false)?;
 
         headers.insert(
             HeaderName::from_static("public-sharing-level"),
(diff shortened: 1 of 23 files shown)


src/client/http.rs
```diff
@@ -34,6 +34,15 @@ impl AybClient {
         }
     }
 
+    fn add_optional_bearer_token(&self, headers: &mut HeaderMap) {
```
marcua: can we merge this with add_bearer_token and pass it an `optional` boolean that defaults to true?

src/client/http.rs
```diff
@@ -34,6 +34,15 @@ impl AybClient {
         }
     }
```
marcua: Sanity-check: Nothing to update in client-js/ayb.js right?

src/server/api_endpoints/database_details.rs
```diff
@@ -14,18 +14,29 @@ pub async fn database_details(
...
+
+    if discoverable {
+        let (can_manage, access_level) = match authenticated_entity.as_ref() {
+            Some(entity) => {
+                let can_manage = can_manage_database(entity, &database, &ayb_db).await?;
+                let access_level =
+                    highest_query_access_level(entity, &database, None, &ayb_db).await?;
+                (can_manage, access_level)
+            }
+            None => (false, None),
```
marcua, line 41: sanity-check: could highest_query_access_level already return None or did the type change?

src/server/api_endpoints/database_details.rs
```diff
@@ -41,11 +52,13 @@ pub async fn database_details(
...
+        let message = match authenticated_entity.as_ref() {
+            Some(entity) => format!(
                 "Authenticated entity {} can't access database {}/{}",
-                authenticated_entity.slug, entity_slug, database_slug
+                entity.slug, entity_slug, database_slug
             ),
-        })
+            None => format!("Database {entity_slug}/{database_slug} is not accessible"),
+        };
+        Err(AybError::Other { message })
```
marcua: Good use for AybError::Unauthorized or no?

src/client/http.rs
marcua: Done in 64e51c8 — merged into a single `add_bearer_token(&mut headers, optional: bool) -> Result<(), AybError>`. Rust doesn't have default arg values, so every caller passes the bool explicitly: `false` for the auth-required endpoints, `true` for `entity_details` / `database_details`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/client/http.rs
marcua: Confirmed nothing to update in `client-js/ayb.js`. That library only exposes `query` (and OAuth/config helpers) — there's no wrapper for `entity_details` or `database_details`, so opening up anonymous access on the Rust side doesn't change any JS surface. Auth is handled inline in `query()` (line 156: `'Authorization': 'Bearer ${token}'`), there's no `add_bearer_token`-equivalent helper to merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/api_endpoints/database_details.rs
marcua: Type unchanged. `highest_query_access_level` returns `Result<Option<QueryMode>, AybError>` (`src/server/permissions.rs:116`) and `DatabaseDetails.highest_query_access_level` is `Option<QueryMode>` (`src/http/structs.rs:201`). Setting it to `None` for anon viewers fits the existing shape — no migration on the field, just an additional case where it's null.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/api_endpoints/database_details.rs
marcua: Yes, good call — switched to `AybError::Unauthorized` in 64e51c8. The "not discoverable" path now returns 401 for both the anon "log in to access" case and the auth "you don't have access" case, instead of the misleading 500 the `Other` variant produced. Existing tests that check the error message text still pass; only the status code changes. (Strict 403/404-on-private semantics would be a follow-up — the existing TODO at the top of the function flags the 404-to-not-leak-existence question.)

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/server_runner.rs
```diff
@@ -26,10 +26,18 @@ pub fn config(cfg: &mut web::ServiceConfig, ayb_config: &AybConfig) {
...
+    // requests through without an `Authorization` header. Endpoints that
+    // require authentication enforce that themselves via
+    // `unwrap_authenticated_entity`, which returns 401 when no entity is
+    // attached. Endpoints that allow anonymous read access to publicly-
+    // shared resources (entity_details, database_details) inspect the
+    // optional entity directly.
     cfg.service(
         web::scope("/v1")
-            .wrap(HttpAuthentication::bearer(entity_validator))
+            .wrap(middleware::from_fn(optional_entity_validator))
```
marcua: I had expected to keep             .wrap(HttpAuthentication::bearer(entity_validator)) for all but the two relevant endpoints, which would get middleware::from_fn(optional_entity_validator). Is that feasible?

src/server/server_runner.rs
```diff
@@ -87,36 +95,48 @@ pub fn config(cfg: &mut web::ServiceConfig, ayb_config: &AybConfig) {
     }
 }
 
-async fn entity_validator(
+async fn optional_entity_validator(
```
marcua: Didn't review this yet, will await decision on comment above

src/server/ui_endpoints/templates/database.html
```diff
@@ -103,7 +103,14 @@ <h3 class="text-lg font-medium mb-2">Database querying</h3>
                 {% else %}
                     <div class="uk-alert uk-alert-destructive" data-uk-alert="">
                         <div class="uk-alert-title">You don't have query access to this database.</div>
+                        {% if logged_in_entity %}
                         <p>You can request access from the database owner or fork a copy.</p>
+                        {% else %}
+                        <p>
+                            <a href="/log_in?next=/{{ entity }}/{{ database }}" class="underline">Log in</a>
+                            to query this database, or request access from the database owner.
```
marcua:
```suggestion
                            <a href="/log_in?next=/{{ entity }}/{{ database }}" class="underline">Login</a>
                            is required for query access.
```

tests/e2e_tests/anonymous_access_tests.rs
```diff
@@ -0,0 +1,158 @@
+use crate::e2e_tests::{FIRST_ENTITY_DB, FIRST_ENTITY_DB2, FIRST_ENTITY_SLUG};
```
marcua, line 1: Note to self: haven't reviewed yet

src/server/server_runner.rs
marcua: Yes, feasible — implemented in b8a133f. The "two sibling `web::scope("/v1")` blocks, one per middleware" form I'd originally drafted didn't actually work: actix-web doesn't fall through between scopes that share a path prefix; the first scope owns the prefix and a non-matching inner route just 404s without trying the next scope (this is what bit me on the first push and forced the centralized handler-level enforcement).

The shape we wanted is reachable a different way: `wrap = ...` on the per-endpoint route macro. So the /v1 scope itself has no `wrap`, and:
  - 10 auth-required endpoints declare `wrap = "HttpAuthentication::bearer(entity_validator)"` — middleware rejects missing/invalid bearer at the routing layer, exactly as before.
  - `entity_details` and `database_details` declare `wrap = "middleware::from_fn(optional_entity_validator)"` — anon passes through, bad tokens still rejected.

`entity_validator` and `optional_entity_validator` are now `pub` in `server_runner.rs` so the macros can name them; the wrap expressions use fully-qualified paths so endpoint files don't need extra imports. Tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/ui_endpoints/templates/database.html
marcua: Applied in b8a133f.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/server_runner.rs
```diff
@@ -87,36 +101,78 @@ pub fn config(cfg: &mut web::ServiceConfig, ayb_config: &AybConfig) {
...
+            }
+        };
+
+        let api_token = retrieve_and_validate_api_token(&token, &ayb_db).await?;
+        let entity = ayb_db.get_entity_by_id(api_token.entity_id).await?;
+        req.extensions_mut().insert(entity);
+        req.extensions_mut().insert(api_token);
     }
+
+    next.call(req).await
```
marcua, line 148: What's this line?

src/server/server_runner.rs
```diff
@@ -87,36 +101,78 @@ pub fn config(cfg: &mut web::ServiceConfig, ayb_config: &AybConfig) {
...
+                    message: "Misconfigured server: no database".to_string(),
+                }
+                .into());
+            }
+        };
+
+        let api_token = retrieve_and_validate_api_token(&token, &ayb_db).await?;
+        let entity = ayb_db.get_entity_by_id(api_token.entity_id).await?;
+        req.extensions_mut().insert(entity);
+        req.extensions_mut().insert(api_token);
```
marcua: This stuff looks very similar to the code in entity_validator. Any way to share?

src/server/server_runner.rs
```diff
@@ -26,10 +27,23 @@ pub fn config(cfg: &mut web::ServiceConfig, ayb_config: &AybConfig) {
...
+    //     rejected at the middleware layer.
+    //   - `middleware::from_fn(optional_entity_validator)` for the two
+    //     anonymous-readable endpoints (`entity_details`,
+    //     `database_details`) — anonymous requests pass through, but
+    //     invalid / revoked / expired bearer tokens are still rejected (a
+    //     bad token is never silently downgraded to anonymous).
+    // The /v1 scope itself has no `wrap`; per-endpoint wraps stack on the
+    // resource directly. This is the only way to mix two middlewares on a
+    // shared path prefix in actix-web — sibling scopes at `/v1` don't
+    // fall through.
```
marcua:
```suggestion
```

marcua: Done in 78f5565 — paragraph removed.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/server_runner.rs
marcua: Good call, deduped in 78f5565. Extracted `attach_validated_entity(token, &req)` (the validate-token + lookup-entity + insert-into-extensions sequence) and reduced both validators to thin wrappers. The only remaining difference is the actix-web error shape each one has to return (`(Error, ServiceRequest)` for the bearer extractor vs. plain `Error` for `from_fn`) and the optional middleware's branch on the absence of an `Authorization` header.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/server/server_runner.rs
marcua: `next.call(req).await` is the call into the next service in the middleware chain — typically the actual route handler. This is the required signature for `actix_web::middleware::from_fn` middleware: the function receives the `ServiceRequest` and a `Next<B>`, and is responsible for invoking `next.call(req)` exactly once to forward to whatever's downstream (or returning early without calling `next` to short-circuit, e.g. for an auth rejection).

Without this line, requests that reach `optional_entity_validator` would never get to the route handler — the middleware would just return early and the response would be empty / hung. We don't short-circuit anywhere in the success path; only the bad-bearer-token case returns `Err` instead of calling `next`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/e2e_tests/anonymous_access_tests.rs
```diff
@@ -0,0 +1,160 @@
...
+    // Make test.sqlite publicly readable so anonymous viewers can discover it.
+    update_database(
+        config_path,
+        first_token,
+        FIRST_ENTITY_DB,
+        Some("read-only"),
+        "Database e2e-first/test.sqlite updated successfully",
+    )?;
+
+    let client = reqwest::Client::builder()
```
marcua: If it's not too convoluted, consider using existing client convenience methods (in crate::utils::ayb) throughout this file so we test from the CLI all the way to the endpoint.

tests/e2e_tests/anonymous_access_tests.rs
marcua: Routed most assertions through the CLI in 30002e5. Added two helpers in `tests/utils/ayb.rs`:

- `list_databases_no_auth(config, server_url, entity, format, result)`
- `database_details_no_auth(config, server_url, database, result)`

Both take a tempfile-backed config path so the CLI doesn't pick up cached tokens for `server_url` from the main test config, and pass `--url` explicitly. The CLI may write `default_url` into the tempfile on first invocation but never any auth, so subsequent calls in the same test reuse it cleanly. The bad-bearer-token assertion uses the existing `list_databases` helper with a corrupted real token — exercises the server middleware end-to-end and surfaces the same `Error: Invalid API token` text the existing tests assert.

Two checks stay on raw reqwest: the anonymous query/create rejections. The CLI's `add_bearer_token(optional=false)` short-circuits client-side when no token is set ("Calling endpoint that requires client API token, but none provided"), so the CLI can't actually send a no-`Authorization`-header request to those endpoints; the only way to exercise the server-side 401 is to bypass the client. I left a comment in the test explaining why those two cases are different.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #776, opened 2026-05-25
DuckDB support

marcua:
## DuckDB support

Adds DuckDB as a second hosted-database engine alongside SQLite. Users pick the engine at database-creation time; everything downstream (query execution, the sandboxed query daemon, snapshots/restore, the web UI, and the CLI) works for both.

```bash
ayb client create_database entity/mydb.duckdb duckdb
```

### What's included

- **Engine abstraction.** A `DbEngine` trait (`src/hosted_db/engine.rs`) with `SqliteEngine` and `DuckdbEngine` implementations. The query daemon and snapshot creation dispatch through `engine_for(&DBType)` instead of scattered `match db_type` arms. Both engines are stateless zero-sized types, so `engine_for` hands out a shared `&'static dyn DbEngine` rather than allocating per call. `DBType` flows through as an enum end-to-end; it only becomes a string at the daemon's `argv` boundary.
- **Trait surface is safe by construction.** `DbEngine::query` takes no `allow_unsafe` flag — every call runs the engine's full safety perimeter (no extension loading, no external file/network access, no ATTACH). Engines relax that internally only where an operation requires it (snapshots need ATTACH), as an implementation detail callers can't reach.
- **Query daemon** is db-type aware (`ayb_query_daemon <path> <db_type>`) and executes against the right engine.
- **Snapshots/restore** for DuckDB via `ATTACH … COPY FROM DATABASE`, paralleling SQLite's `VACUUM INTO`. The per-engine `SqliteSnapshotMethod`/`DuckdbSnapshotMethod` config enums were removed — each engine has one method, so they were dead config.
- **Web UI** gains a SQLite/DuckDB selector on the create-database form.
- **DuckDB is bundled** (compiled from source, like `rusqlite`), so the binary stays self-contained with no system dependency.
- **Tests:** DuckDB create/query and snapshot/restore coverage at the API layer, a DuckDB browser flow (create via the web form, insert, query, snapshot, restore) mirroring the SQLite UI tests, and unit tests for value conversion, error classification, and slug validation.

### Security-relevant changes

- **Slug validation at the API boundary.** Entity and database slugs are now validated server-side (`src/server/validation.rs`): non-empty, not `.`/`..`, and restricted to `[A-Za-z0-9._-]`. Previously the only check was the `pattern` attribute on the web form, which is trivially bypassed by calling the API directly. Slugs become filesystem path components, so `..` and separators were a traversal vector. New `AybError::InvalidSlug` variant.
- **Escaped SQL string literals.** Snapshot statements interpolate paths (which embed user-controlled slugs) into SQL. Both engines now render them through `sql_string_literal`, which single-quotes and doubles embedded quotes. This matters because snapshots run in the **server process**, not in the sandboxed daemon. SQLite's `VACUUM INTO` also moves from double quotes (an identifier in SQL) to a proper string literal.
- **Merged validation modules.** `slug_validation` and `username_validation` are now one `server::validation` module with named entry points (`validate_entity_slug`, `validate_database_slug`) instead of a stringly-typed `kind` argument. Reserved names apply to entities (which sit at the root of the URL space) but not to databases (which are namespaced under an entity).
- **Cleared daemon environment.** `build_daemon_command` calls `env_clear()`, so the daemon inherits none of the server's secrets (fernet key, S3 credentials, SMTP password) — they can't be read from `/proc/self/environ` even if a query escaped the engine's file-access restrictions.

### Engine differences worth knowing

- **INSERT results differ.** DuckDB returns a one-row `Count` result (affected rows) for `INSERT`/`UPDATE`/`DELETE`; SQLite returns no rows. DDL such as `CREATE TABLE` returns a `Count` column with zero rows. This is surfaced as-is rather than normalized ([DuckDB docs](https://duckdb.org/docs/current/sql/statements/insert)).
- **File locking.** DuckDB takes a whole-file lock on open and fails immediately on conflict, where SQLite waits out contention via `busy_timeout`. Since the snapshot job opens each database on a timer, `open_with_retry`/`with_lock_retry` poll for up to 5 seconds (matching SQLite's `busy_timeout`) before giving up. The `duckdb` crate exposes no structured code for this — every failure carries `ErrorCode::Unknown` — so classification matches on message text, pinned by unit tests that provoke a real lock conflict and a real read-only violation so a crate upgrade that rewords them fails loudly.
- **Sandbox.** The query daemon's Landlock + setrlimit sandbox uses one set of limits for both engines (256 MB address space / 256 MB file size / 32 fds). DuckDB needs more than SQLite's old 64 MB — it can't even initialize under it. DuckDB also reads `/proc/self` and `/sys` at startup to size its thread pool and buffer pool, so those are allowed read-only; `Config.threads(1)` + `max_memory("128MB")` keep it inside the budget regardless of host size.

### Tradeoffs / things to weigh before merge

- **Build time.** Bundled DuckDB compiles a large C++ amalgamation. A cold CI build (build + clippy) runs ~25 minutes; warm runs with a valid `rust-cache` are ~3–4 minutes end to end, since `libduckdb-sys` only recompiles when its version, the toolchain, or profile settings change. `cache-on-failure: "true"` keeps red iterations from recompiling it.
- **Dev-profile override.** `[profile.dev.package.libduckdb-sys]` sets `debug = 0` and `opt-level = 2`. Debug artifacts for the amalgamation dominated disk usage (~13 GB → ~1.9 GB for the whole `target/`), and tests now exercise an optimized DuckDB instead of an `-O0` one. Trade-off: no debuggable frames inside DuckDB itself; Rust code is unaffected, as are release builds.
- **`cargo install` cost.** End users compile DuckDB once per install/upgrade. Switching to `DUCKDB_DOWNLOAD_LIB`/system linking would speed builds but produces a dynamically-linked binary whose `.so` isn't installed alongside it (dangling rpath) — strictly worse for `cargo install`. If fast installs matter later, the right lever is prebuilt release binaries (via the existing `publish-on-tag` workflow / `cargo-binstall`), not changing the link mode. A feature flag could also let SQLite-only users skip the DuckDB compile.

### Verification

- README's create/list flow was replicated against a real server (register → confirm → create SQLite → create DuckDB → list → query both). The documented `ayb client list` output, including row order and column widths, matches actual output.
- Full e2e suite green, including the browser tests and the DuckDB screenshot baselines.

### Follow-ups (not in this PR)

- #798 — run snapshots inside the isolation boundary (they currently parse tenant database files in the server process), and the persistent-connection handoff that unlocks.
- #799 — lint test code (`cargo clippy --all-targets`); `tests/` is currently unlinted.
- #810 — flaky browser test from overlapping htmx query submits (~7%, pre-existing, in the SQLite snapshot test).
- #811 — cap slug length so long names fail validation rather than at filesystem `ENAMETOOLONG`.
- #812 — strip bidi/invisible Unicode controls from entity profile text.
- Optional `duckdb` feature flag for SQLite-only builds.
- Prebuilt release binaries for fast installs.


src/hosted_db.rs
```diff
@@ -77,17 +81,22 @@ impl TabularFormatter for QueryResult {
     }
 }
 
+pub fn engine_for(db_type: &DBType) -> Box<dyn DbEngine> {
```
marcua: Here it's DBType, but isn't the daemon registry and query daemon passing around a db_type string? (I prefer the enum, just trying to understand how types align)

src/hosted_db/sandbox.rs
```diff
@@ -103,9 +103,8 @@ fn print_warning_banner(details: &[&str]) {
 /// - Filesystem: only the database file (read-write) and shared
 ///   libraries (read-only) are accessible.
 /// - Network: all TCP bind/connect denied (on kernel 6.7+).
-/// - Memory: 64 MB virtual memory limit (RLIMIT_AS).
-/// - File size: 75 MB max file size (RLIMIT_FSIZE).
-/// - File descriptors: 10 max open files (RLIMIT_NOFILE).
+/// - SQLite limits: 64 MB memory, 75 MB file size, 10 file descriptors.
```
marcua: Let's make both DBs require the same limits (so if DuckDB is more consumptive, use its minimums). Hopefully don't need to pass a type in at all at that point.

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,227 @@
...
+            AybError::NoWriteAccessError {
+                message: "Attempted to write to database while in read-only mode".to_string(),
+            }
+        }
+        _ => AybError::from(err),
+    }
+}
+
+fn duckdb_value_to_string(value: Value) -> Option<String> {
+    match value {
```
marcua, line 217: Looking at this more exhaustive list in DuckDB type mappings, are we missing any in SQLite?

src/hosted_db/sqlite.rs
```diff
@@ -1,11 +1,75 @@
...
+                message: format!("Snapshot failed integrity check: {result:?}"),
+            });
+        }
+        Ok(())
+    }
+
+    fn db_type_str(&self) -> &'static str {
+        "sqlite"
+    }
+}
```
marcua, line 51: Should query_sqlite still be pub? (Same for the equivalent DuckDB function)

src/hosted_db/sqlite.rs
```diff
@@ -1,11 +1,75 @@
...
+        snapshot_path: &Path,
+    ) -> Result<(), AybError> {
+        let backup_query = match config.sqlite_method {
+            SqliteSnapshotMethod::Backup => {
+                return Err(AybError::SnapshotError {
+                    message: "Backup requires dot commands, which are not yet supported"
+                        .to_string(),
+                })
+            }
+            SqliteSnapshotMethod::Vacuum => {
```
marcua: Since we only really have Vacuum here and one method on DuckDB, let's remove the config settings (and update README where necessary), and only use our one backup method for each DB type.

src/hosted_db/sqlite.rs
```diff
@@ -1,11 +1,75 @@
...
+        {
+            return Err(AybError::SnapshotError {
+                message: format!("Snapshot failed integrity check: {result:?}"),
+            });
+        }
+        Ok(())
+    }
+
+    fn db_type_str(&self) -> &'static str {
+        "sqlite"
```
marcua: Related to another comment, I broadly don't understand why we use strings here and not the enum type

tests/utils/ayb.rs
```diff
@@ -28,7 +28,17 @@ pub fn create_database(
     database: &str,
     result: &str,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let cmd = ayb_assert_cmd!("client", "--config", config, "create_database", database, "sqlite"; {
+    create_database_with_type(config, api_key, database, "sqlite", result)
+}
+
+pub fn create_database_with_type(
```
marcua: Can we add db_type to the create_database function and update it everywhere rather than having two helper functions?

tests/e2e_tests/create_and_query_db_tests.rs
```diff
@@ -164,3 +168,43 @@ pub fn test_create_and_query_db(
...
+        &api_keys.get("first").unwrap()[0],
+        FIRST_ENTITY_DUCKDB,
+        "duckdb",
+        "Successfully created e2e-first/test.duckdb",
+    )?;
+
+    query(
+        config_path,
+        &api_keys.get("first").unwrap()[0],
+        "CREATE TABLE duck_test(id INTEGER, name VARCHAR);",
```
marcua: same test table and values and test query as we use for the sqlite version

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,227 @@
...
+    fn create_snapshot(
+        &self,
+        config: &AybConfigSnapshots,
+        db_path: &Path,
+        snapshot_path: &Path,
+    ) -> Result<(), AybError> {
+        match config.duckdb_method {
+            DuckdbSnapshotMethod::CopyDatabase => {
+                let copy_query = format!(
+                    "ATTACH '{}' AS snapshot_dest;COPY FROM DATABASE main TO snapshot_dest;",
```
marcua: 1) Any integrity checks to run like we do in SQLite after the backup is done? (Don't make them up, only if there's an equivalent in DuckDB.)
2) Any considerations for concurrent queries while backup is running? Will we get a consistent snapshot? Will we harm other queries? I know there are vacuum tradeoffs with SQLite, and want to understand the DuckDB backup tradeoffs and list of backup methods.

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,227 @@
...
+        &self,
+        path: &Path,
+        query: &str,
+        allow_unsafe: bool,
+        query_mode: QueryMode,
+    ) -> Result<QueryResult, AybError> {
+        query_duckdb(&path.to_path_buf(), query, allow_unsafe, query_mode)
+    }
+
+    fn create_snapshot(
```
marcua: Can we add a simple update/snapshot/revert test to e2e paralleling the sqlite one? All good if it's simpler.

src/server/ui_endpoints/templates/create_database_fields.html
```diff
@@ -10,6 +10,27 @@
         title="Only letters, numbers, underscores, hyphens, and periods are allowed">
 </div>
 
+<div class="mb-4">
```
marcua, line 16: do our UI e2e tests need to update screenshots for this?

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+pub fn query_duckdb(
+    path: &PathBuf,
+    query: &str,
+    allow_unsafe: bool,
+    query_mode: QueryMode,
+) -> Result<QueryResult, AybError> {
+    let config = duckdb::Config::default()
+        .access_mode(match query_mode {
```
marcua: Can e2e tests ensure that read-only vs readwrite works (like we do for SQLite)?

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,227 @@
...
+        })
+        .map_err(|e| AybError::Other {
+            message: format!("DuckDB config error: {e}"),
+        })?;
+
+    let conn = duckdb::Connection::open_with_flags(path, config)?;
+
+    if !allow_unsafe {
+        conn.execute_batch(
+            "SET autoinstall_known_extensions=false;
```
marcua, line 62: How's this compare to the SQLite equivalents---anything missing, anything overboard, etc.? What does each flag do?

.github/workflows/tests.yml
```diff
@@ -40,6 +40,12 @@ jobs:
         # GitHub automatically evicts caches after 7 days of inactivity
         # rust-cache also cleans build artifacts older than 1 week
         shared-key: "ayb-tests"
+        # Save the cache even when the job fails. The bundled DuckDB C++
+        # build is large (~13 GB of debug artifacts, 20+ min cold) and by
+        # default the cache is only saved on a successful job -- so while a
+        # branch iterates on a red build, every run recompiles DuckDB from
+        # scratch. Saving on failure lets subsequent runs reuse it.
```
marcua:
```suggestion
```

follow-ups.md
```diff
@@ -0,0 +1,167 @@
+# Follow-ups
```
marcua: to-do: copy from here and delete before merging

src/hosted_db.rs
```diff
@@ -77,17 +81,44 @@ impl TabularFormatter for QueryResult {
...
+pub(crate) fn sql_string_literal(path: &Path) -> String {
+    format!("'{}'", path.display().to_string().replace('\'', "''"))
+}
+
+/// Return the engine for `db_type`.
+///
+/// Both engines are zero-sized structs with no per-database state, so a
+/// single shared instance of each serves every caller; returning a
+/// `&'static dyn` keeps the dynamic dispatch while avoiding an
+/// allocation on every query and snapshot.
```
marcua:
```suggestion
```

src/hosted_db.rs
```diff
@@ -77,17 +81,44 @@ impl TabularFormatter for QueryResult {
...
+pub fn engine_for(db_type: &DBType) -> &'static dyn DbEngine {
+    match db_type {
+        DBType::Sqlite => &SqliteEngine,
+        DBType::Duckdb => &DuckdbEngine,
+    }
+}
+
 pub async fn run_query(
```
marcua: What's the benefit of this function any more given a caller has a daemon registry and can call execute_query directly?

src/hosted_db/daemon_registry.rs
```diff
@@ -114,25 +116,42 @@ impl DaemonRegistry {
...
+        let mut cmd = build_daemon_command(db_path, db_type)?;
+
+        // Run the query daemon with an empty environment. It inherits
+        // nothing from the server process, so server secrets (fernet key,
+        // S3 credentials, SMTP password) never reach the sandboxed daemon
+        // -- they can't be read out of /proc/self/environ even if a query
+        // somehow escaped DuckDB's file-access restrictions. The daemon
+        // needs no environment: it is invoked by absolute path, links only
+        // standard system libraries, and resource/sandbox limits come from
+        // setrlimit/Landlock rather than env vars.
```
marcua:
```suggestion
```

src/hosted_db/daemon_registry.rs
```diff
@@ -114,25 +116,42 @@ impl DaemonRegistry {
...
+        cmd.env_clear();
```
marcua: should env_clear live in build_daemon_command, or good here?

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,414 @@
...
+        // catalog is named after the file, not "main". Opening the source
+        // read-only would also make the whole instance (including the
+        // attached destination) read-only, breaking the COPY. With an
+        // in-memory read-write main, the source is attached READ_ONLY (so
+        // it is never modified) and the destination read-write.
+        //
+        // Both paths are rendered as escaped SQL string literals: they
+        // embed user-controlled entity and database slugs, and this
+        // statement runs in the server process rather than in a
+        // sandboxed daemon.
```
marcua:
```suggestion

```

src/hosted_db/duckdb.rs
```diff
@@ -0,0 +1,414 @@
...
+    };
+
+    let mut results: Vec<Vec<Option<String>>> = Vec::new();
+    while let Some(row) = rows.next().map_err(map_duckdb_error)? {
+        let mut result: Vec<Option<String>> = Vec::new();
+        for col_idx in 0..num_columns {
+            let value: duckdb::Result<Value> = row.get(col_idx);
+            match value {
+                Ok(val) => result.push(duckdb_value_to_string(val)),
+                Err(_) => result.push(None),
```
marcua: Does this silently ignore errors?

src/hosted_db/engine.rs
```diff
@@ -0,0 +1,25 @@
...
+use std::path::Path;
+
+/// A hosted database engine (SQLite or DuckDB).
+///
+/// Implementations are stateless; `engine_for` hands out a shared
+/// instance of each. Every method runs the engine's safety perimeter
+/// (no extension loading, no external file/network access, no ATTACH),
+/// so callers cannot ask for an unrestricted connection. Engines relax
+/// those restrictions internally where a specific operation requires it
+/// -- snapshots need ATTACH -- but that stays an implementation detail
```
marcua:
```suggestion
/// , e.g., snapshots need ATTACH, but that stays an implementation detail
```

src/server/slug_validation.rs
```diff
@@ -0,0 +1,79 @@
...
+///   the server process rather than in a sandboxed daemon, so it is
+///   worth rejecting these at the boundary rather than relying on a
+///   single layer.
+///
+/// The permitted set matches the `pattern` attribute on the
+/// create-database form, which previously existed only in the browser
+/// and so was trivially bypassed by calling the API directly.
+pub fn validate_slug(kind: &str, slug: &str) -> Result<(), AybError> {
+    if slug.is_empty() {
+        return Err(AybError::Other {
```
marcua: Can we use or create a more specific AybError:: type?

tests/browser_e2e_tests/create_and_query_database.rs
```diff
@@ -105,9 +104,15 @@ pub async fn test_create_and_query_database_flow(
         .click(Some(ClickOptions::builder().timeout(5000.0).build()))
         .await?;
 
+    // Let the first insert's request finish before submitting the next
+    // one. The query form posts via htmx with no hx-sync, so two submits
+    // fired back-to-back from the same form overlap and one of the
+    // inserts is lost. Every other step here is separated by a
+    // screenshot_compare (which settles briefly); these two are not.
```
marcua, line 109:
```suggestion
    // Let the first insert's request finish before submitting the next
    // one. Every other step here is separated by a
    // screenshot_compare (which settles briefly); these two are not.
```

tests/browser_e2e_tests/duckdb.rs
```diff
@@ -0,0 +1,352 @@
...
+        .await
+        .first()
+        .click(Some(ClickOptions::builder().timeout(5000.0).build()))
+        .await?;
+
+    // Let the first insert's request finish before submitting the next
+    // one. The query form posts via htmx with no hx-sync, so two submits
+    // fired back-to-back from the same form overlap and one of the
+    // inserts is lost. Every other step here is separated by a
+    // screenshot_compare (which settles briefly); these two are not.
```
marcua, line 122:
```suggestion
    // Let the first insert's request finish before submitting the next
    // one. Every other step here is separated by a
    // screenshot_compare (which settles briefly); these two are not.
```

tests/e2e.rs
```diff
@@ -97,6 +98,11 @@ async fn client_server_integration(
     test_anonymous_access(&config_path, &api_keys, server_url).await?;
     test_token_management(&config_path, &api_keys)?;
     test_oauth_token_exchange_errors(server_url).await?;
+    // Runs last: it creates a DuckDB database under e2e-first, which would
+    // otherwise appear in the many exact database-listing assertions above
+    // (and in the periodic snapshot job during test_snapshots).
```
marcua:
```suggestion
```

tests/e2e.rs
```diff
@@ -152,5 +158,9 @@ async fn browser_e2e() -> Result<(), Box<dyn std::error::Error>> {
     // Test token management UI (uses the OAuth read-only token for revocation testing)
     test_token_management_flow(&page, &username, &base_url, readonly_token).await?;
 
+    // Runs last: creates a DuckDB database whose periodic snapshots would
+    // otherwise perturb the SQLite snapshot-count assertions above.
```
marcua:
```suggestion
    // Test DuckDB flows
```

tests/e2e_tests/snapshot_tests.rs
```diff
@@ -297,11 +300,97 @@ pub async fn test_snapshots(
...
+pub async fn test_snapshots_duckdb(
+    db_type: &str,
+    config_path: &str,
+    api_keys: &HashMap<String, Vec<String>>,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let api_key = &api_keys.get("first").unwrap()[0];
+
+    // Object storage (MinIO) persists across test runs even though the
+    // data directory is reset, so clear any DuckDB snapshots left behind
+    // by previous runs to get a deterministic starting count.
```
marcua:
```suggestion
    // Remove all snapshots so our tests aren't affected by
    // timing/snapshots from previous tests.
```

README.md
```diff
@@ -99,13 +99,13 @@ Rows: 0
 # If you don't pass a query to the query command, ayb launches an interactive query session
```
marcua, line 104: Add short mention of DuckDB, and double-check that no commands need update to make sqlite explicit. (If not strictly required to update command, can skip.)

src/hosted_db/engine.rs
```diff
@@ -0,0 +1,25 @@
...
+use std::path::Path;
+
+/// A hosted database engine (SQLite or DuckDB).
+///
+/// Implementations are stateless; `engine_for` hands out a shared
+/// instance of each. Every method runs the engine's safety perimeter
+/// (no extension loading, no external file/network access, no ATTACH),
+/// so callers cannot ask for an unrestricted connection. Engines relax
+/// those restrictions internally where a specific operation requires it
+/// , e.g., snapshots need ATTACH, but that stays an implementation detail
```
marcua:
```suggestion
/// those restrictions internally where a specific operation requires it
/// (e.g., snapshots need ATTACH) but that stays an implementation detail
```

src/server/validation.rs
```diff
@@ -0,0 +1,147 @@
...
+/// attribute is a browser-side convenience only; this function is the
+/// authoritative check, since the API can be called directly.
+fn validate_slug_syntax(kind: &str, slug: &str) -> Result<(), AybError> {
+    if slug.is_empty() {
+        return Err(AybError::InvalidSlug {
+            message: format!("A {kind} slug can't be empty"),
+        });
+    }
+
+    if slug == "." || slug == ".." {
```
marcua, line 55: I'm curious why regex on frontend and do character/string comparisons on backend. Are the two equivalent? Is one preferable? Don't change yet, just answer

PR #777, opened 2026-05-25
Fix flaky snapshot test

marcua:
No more hard-coded sleep---explicitly wait for snapshot count or latest snapshot ID to change, with a 20s timeout.

### DIFF
--- a/tests/e2e_tests/snapshot_tests.rs
+++ b/tests/e2e_tests/snapshot_tests.rs
@@ -33,6 +33,34 @@ fn wait_for_snapshot_count(
     }
 }
 
+/// Wait until the newest snapshot ID differs from `prev_newest_id`,
+/// then return the current snapshot list. This is needed for the
+/// pruning test where the count stays the same (6 → 7 → 6) but the
+/// contents change.
+fn wait_for_new_snapshot(
+    config_path: &str,
+    api_key: &str,
+    database: &str,
+    prev_newest_id: &str,
+    timeout_secs: u64,
+) -> Vec<ayb::server::snapshots::models::ListSnapshotResult> {
+    let deadline = time::Instant::now() + time::Duration::from_secs(timeout_secs);
+    loop {
+        thread::sleep(time::Duration::from_secs(2));
+        let snapshots = list_snapshots(config_path, api_key, database, "csv")
+            .expect("failed to list snapshots");
+        let newest_changed = !snapshots.is_empty() && snapshots[0].snapshot_id != prev_newest_id;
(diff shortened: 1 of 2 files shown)


tests/e2e_tests/snapshot_tests.rs
```diff
@@ -37,36 +89,20 @@ pub async fn test_snapshots(
...
-    assert_eq!(
-        snapshots.len(),
         1,
-        "there should be one snapshot after sleeping"
+        20,
     );
+
+    let last_modified_at = snapshots[0].last_modified_at;
     // No change to database, so same number of snapshots after sleep.
     thread::sleep(time::Duration::from_secs(4));
```
marcua, line 96: How are we picking when to sleep/list vs use the wait function? Are we being inconsistent anywhere?

tests/e2e_tests/snapshot_tests.rs
```diff
@@ -37,36 +89,20 @@ pub async fn test_snapshots(
...
         FIRST_ENTITY_DB,
-        "csv",
-    )?;
-
-    let last_modified_at = snapshots[0].last_modified_at;
-    assert_eq!(
-        snapshots.len(),
         1,
-        "there should be one snapshot after sleeping"
+        20,
```
marcua: We're always passing in 20, so can we make that the default?

tests/e2e_tests/snapshot_tests.rs
```diff
@@ -1,12 +1,64 @@
-use crate::e2e_tests::{
-    FIRST_ENTITY_DB, FIRST_ENTITY_DB_CASED, FIRST_ENTITY_DB_SLUG, FIRST_ENTITY_SLUG,
-};
+use crate::e2e_tests::{FIRST_ENTITY_DB, FIRST_ENTITY_DB_SLUG, FIRST_ENTITY_SLUG};
```
marcua, line 1: Are we using the old sleep approach in any other test, like the ui tests?

tests/e2e_tests/snapshot_tests.rs
```diff
@@ -1,12 +1,64 @@
...
+            return snapshots;
+        }
+    }
+}
+
+/// Wait until the newest snapshot ID differs from `prev_newest_id`,
+/// then return the current snapshot list. This is needed for the
+/// pruning test where the count stays the same (6 → 7 → 6) but the
+/// contents change.
+fn wait_for_new_snapshot(
```
marcua: We only use this once. Any way to fold it into wait_for_snapshot_count without getting too hacky? List options.

